PR #403, opened 2023-12-15
Fix typo in license

AaronDewes:
It looks like you mistakenly called your software "Open Source". 

However, to be open source, according to opensource.org/osd, section 5:

The license must not discriminate against any person or group of persons. As this license does so, it is not open source.


aaronisme:
@AaronDewes License has been updated

PR #514, opened 2024-01-20
FEAT: Support KUJI

GiMa-Maya:
## Pre-merge check list


GiMa-Maya:
Looks like it’s all good, or am I missing something?

soralit:
Thanks for your PR @GiMa-Maya . We're researching Cosmos chains these days to determine whether we should support it. Will keep you updated once we integrated Kujira.

PR #633, opened 2024-03-18
feat: change coinType for agoric from 564 to 118

LuqiPan:
## Explanation

First of all, thank you for supporting Agoric chain on Keystone.

When I started using my keystone device, I noticed that the address for Agoric chain on my Keystone 3 Pro is different from the address on my Ledger Nano X, even though they share the same seed phrases.

Upon some digging, it seems like in this situation the coinType should be 118 instead of 564. Here are some relevant comments I found:

> My rough understanding is that 564 is a coin type we registered, but it's necessary to use 118, the default cosmos coinType, in certain interop situations; in particular, when using a ledger.

From: https://github.com/Agoric/agoric-sdk/discussions/5830#discussioncomment-5071717

PS: let me know if there are other places where I should update the coinType

## Pre-merge check list

Remark: could you point me to some documentations on how to run tests on a Mac with Apple Silicon?

## How to test

Not sure what's the best way to test this. I'm also not sure if the keystone device would cache the address generated with cointype 564 somewhere?

### DIFF
--- a/rust/apps/cosmos/src/lib.rs
+++ b/rust/apps/cosmos/src/lib.rs
@@ -174,9 +174,9 @@ mod tests {
         }
         {
             //general address BLD-0
-            let root_path = "44'/564'/0'";
+            let root_path = "44'/118'/0'";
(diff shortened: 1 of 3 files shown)


soralit:
Hi @LuqiPan thanks for contributing. But this will affect existing users' devices, we'd consider it more seriously to do the change.

Issue #1096, opened 2024-05-31
Seeking slip39 tests


onvej-sl:
I'm trying to implement a [recent revision](https://github.com/satoshilabs/slips/commit/8d060706b549af6443e04f55605b71f65c981663) of the [SLIP-39 specification](https://github.com/satoshilabs/slips/blob/master/slip-0039.md).

I would like to ask
  * whether you have tests covering slip39 mnemonic generation and recovery and
  * (if so) what the easiest way is to run them.

aaroninks:
Thanks for the question, 
Do you want to implement this on the Keystone3 firmware?

aaroninks:
The test vector defined in the SLIP-39 file can be helpful. https://github.com/satoshilabs/slips/blob/master/slip-0039.md#test-vectors. We are also using it to test our implementation.

onvej-sl:
> Do you want to implement this on the Keystone3 firmware?

Yes.

> We are also using it to test our implementation.

Where can I find the tests? How do I run them?

aaroninks:
> > Do you want to implement this on the Keystone3 firmware?
> 
> Yes.
> 
> > We are also using it to test our implementation.
> 
> Where can I find the tests? How do I run them?

these test cases are included in another repo which is used internally for testing purpose. The repo is not public for new.

fmunteanu:
@aaroninks I'm unable to import a Trezor backup, probably because latest SLIP-39 is not present into Keystone firmware. See https://github.com/KeystoneHQ/keystone3-firmware/issues/1506.

8de2fdb0:
@aaroninks I refactored the work from @onvej-sl on top of current master. Unfortunately I have the same issues with adding tests.

I see that there are test cases read via USB when the firmware runs on a real device, the simulator doesn't seem to have that ability. Any help with adding test cases would be appreciated.

The changes are fairly small and in theory it should be backward compatible. 
[...]

PR #1102, opened 2024-06-04
Implement slip39 extendable backups

onvej-sl:
This pull request implements a [recent revision](https://github.com/satoshilabs/slips/commit/8d060706b549af6443e04f55605b71f65c981663) of the [SLIP-39 specification](https://github.com/satoshilabs/slips/blob/master/slip-0039.md). The highest bit of the iteration exponent has been repurposed for a new "extendable backup flag". Setting the flag indicates that the random identifier is not used as salt in the encryption of the master secret. This makes it possible to create multiple sets of shares, such that each set of shares uses a different identifier and each set of shares leads to the same master secret for every passphrase. This is a desirable property, which unfortunately was not considered in the initial specification. It allows users to start working with their wallet by creating a single-share (1-of-1) scheme and later upgrade to a multi-share scheme while maintaining the same encrypted master secret and passphrases. Existing SLIP-39 shares remain valid. New shares created using the revised specification use a different checksum customization string and are not compatible with older implementations.

Since I'm not able to run tests (see https://github.com/KeystoneHQ/keystone3-firmware/issues/1096#issuecomment-2145024318), consider this pull request completely untested. Please take extra care to ensure that the changes I made (in particular in `src/managers/account_manager.h`) are backward compatible.


8de2fdb0:
any idea what is needed to continue here ?

onvej-sl:
I did my best to implement slip39 extendable backups. However, since I'm not very familiar with your codebase and am unable to run tests, I'm not certain if it actually works. The next steps would be to:

1. Implement and run unit tests. For test vectors, see [this](https://github.com/satoshilabs/slips/blob/master/slip-0039.md#test-vectors).
2. Conduct a thorough review or test of [this commit](https://github.com/KeystoneHQ/keystone3-firmware/pull/1102/commits/50013e0b760cf9d22d5d3238582f7da4798af16e), as it won't be covered by the unit tests.

fmunteanu:
@onvej-sl can we get the developers attention on this? This must be implemented, we cannot import Trezor single slip39 seed keys into wallet, because of recent changes.

onvej-sl:
> @onvej-sl can we get the developers attention on this?

I'm not sure what you expect from me.

PR #1293, opened 2024-08-22
Add Maya (fork of thorchain)

GiMa-Maya:
Adds MAYAChain (CACAO). Uses thorchain (RUNE) integration code. 

UPDATE: Also added MAYAChain to thorwallet


GiMa-Maya:
Adding Maya to THORWallet integration as well now

PR #1366, opened 2024-09-26
[WIP]Feat/zcash

soralit:
## Keystone Zcash UR Registries

This protocol is based on the [Uniform Resources](https://github.com/BlockchainCommons/Research/blob/master/papers/bcr-2020-005-ur.md). It describes the data schemas (UR Registries) used in Zcash integrations.

### Introduction

Keystone's QR workflow involves two main steps: linking the wallet and signing data, broken down into three sub-steps:

1. **Wallet Linking:** Keystone generates a QR code with public key info for the Watch-Only wallet to scan and import.
2. **Transaction Creation:** The Watch-Only wallet creates a transaction and generates a QR code for Keystone to scan, parse, and display.
3. **Signing Authorization:** Keystone signs the transaction, displays the result as a QR code for the Watch-Only wallet to scan and broadcast.

Two UR Registries are needed for these steps, utilizing the Partially Created Zcash Transaction structure.

#### Unified Full Viewing Key (UFVK)

UFVK is a standard account expression format in Zcash as per [ZIP-316](https://zips.z.cash/zip-0316). It consists of:

1. Transparent
2. Sprout
3. Sapling
4. Orchard

This protocol focuses on the Transparent and Orchard components.

#### CDDL for Zcash Accounts

The specification uses CDDL and includes `crypto-hdkey` and `crypto-key-path` specs defined in https://github.com/BlockchainCommons/Research/blob/master/papers/bcr-2020-007-hdkey.md.

```cddl
zcash-accounts = {
    seed-fingerprint: bytes.32, ; the seed fingerprint specified by ZIP-32 to identify the wallet
    accounts: [+ zcash-ufvk],
    ? origin: text, ; source of data, e.g., Keystone
}

zcash-ufvk = {
    ? transparent: crypto-hdkey,
    orchard: zcash-fvk,
    ? name: text,
}

zcash-fvk = {
    key-path: crypto-key-path,
    key-data: bytes,
}
```

`zcash-ufvk` describes the UFVK of a Zcash account. Each seed has multiple accounts with different indexes. For index 0, `zcash-ufvk` should contain a BIP32 extended public key with path `M/44'/133'/0'` (transparent) and an Orchard FVK with path `M_orchard/32'/133'/0'` (Orchard).

#### CDDL for Zcash PCZT

```cddl
zcash-pczt {
    data: bytes, ; Zcash PCZT, signatures inserted after signing.
}
```


docs/protocols/ur_registrys/draft/zcash.md
```diff
@@ -0,0 +1,59 @@
...
+
+This protocol focuses on the Transparent and Orchard components.
+
+#### CDDL for Zcash Accounts
+
+The specification uses CDDL and includes `crypto-hdkey` and `crypto-key-path` specs defined in https://github.com/BlockchainCommons/Research/blob/master/papers/bcr-2020-007-hdkey.md.
+
+```cddl
+zcash-accounts = {
+    master-fingerprint: uint32, ; the master fingerprint to identify the wallet
```
daira, line 34: ZIP 32 specifies a 16-byte seed fingerprint: https://zips.z.cash/zip-0032#seed-fingerprints
It's a byte array so endian-independent.

docs/protocols/ur_registrys/draft/zcash.md
```diff
@@ -0,0 +1,59 @@
...
+    ? name: text,
+}
+
+zcash-fvk = {
+    key-path: crypto-key-path,
+    key-data: bytes,
+}
+```
+
+`zcash-ufvk` describes the UFVK of a Zcash account. Each seed has multiple accounts with different indexes. For index 0, `zcash-ufvk` should contain a BIP32 extended public key with path `M/44'/133'/0'` (transparent) and an Orchard FVK with path `M_orchard/32'/133'/0'` (Orchard).
```
daira, line 51: Specify that:
* `M` is defined as in https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#master-key-generation
* `M_orchard` (we use $m_{\mathsf{Orchard}}$) is defined as in https://zips.z.cash/zip-0032#orchard-master-key-generation .

`133'` is for mainnet. Like other coins, we use `1'` for testnet [as specified in SLIP 44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md#registered-coin-types).

docs/protocols/ur_registrys/draft/zcash.md
soralit: It's good to use this seed fingerprint instead.

soralit:
An initial idea of the PCZT, modified from https://github.com/str4d/zcash/blob/pczt-wip/src/pczt.proto

```proto
message PartiallyCreatedTransaction {
    // The version of this wire format, for storage.
    uint32 protoVersion = 1;
    PcztGlobal global = 2;
    optional PcztTransparent transparent = 3;
    optional PcztOrchard orchard = 4;
}

message PcztGlobal {
    uint32 txVersion = 1;
    
    //is versionGroupId still needed?
    fixed32 versionGroupId = 2;
    
    uint32 lockTime = 3;
    uint32 expiryHeight = 4;
    uint32 consensusBranchId = 5;
    uint32 networkId = 6;
}

message PcztTransparent {
    repeated PcztTransparentInput inputs = 1;
    repeated PcztTransparentOutput outputs = 2;
}

message PcztTransparentInput{
    bytes pubkey = 1;
    PcztTransparentCoin previousCoin = 2;
    PcztTransparentOutpoint previousOutpoint = 3;
    PcztDerivationPath path = 4;
    repeated bytes signatures = 5;
}

message PcztTransparentCoin {
    uint64 value = 1;
    bytes scriptPubkey = 2;
}

message PcztTransparentOutput{
    PcztTransparentCoin coin = 1;
    
    // if an output has path, it is a change output
    optional PcztDerivationPath path = 2;
}

message PcztTransparentOutpoint {
    bytes txHash = 1;
    uint32 index = 2;
}

message PcztDerivationPath {
    // the 32 bytes seed fingerprint;
    bytes seedFingerprint = 1;
    
    // for orchard it should be pk_d(DiversifiedTransmissionKey)
    // hardware wallet should check this with the derived key from it self.
    bytes pubkey = 2;
    
    repeated PathComponent pathComponents = 3;
}

message PathComponent {
    uint32 value = 1;
    bool harden = 2;
}

message PcztOrchard {
    uint8 flag = 1;
    int64 valueBalance = 2;
    bytes anchor = 3;
    repeated PcztOrchardAction actions = 4;
}

message PcztOrchardAction {
    bytes nf = 1;
    bytes rk = 2;
    bytes cmx = 3;
    //(epk, enc_ciphertext, out_ciphertext)
    bytes encryptedNote = 4;
    bytes cv_net = 5;
    
    //the serilized spend note(rho, rseed, value, receipent) of this action, to verify the nf and cmx in this action.
    bytes spend_note = 6;
    PcztDerivationPath path = 7;
    repeated bytes signatures = 8;
}
```

daira:
```
//is versionGroupId still needed?
```
The purpose of `versionGroupId` is to unambiguously identify the transaction format even if version numbers are reused across Zcash forks, and that's still relevant for PCZTs.

rust/pczt/src/protos/pczt.proto
soralit, line 1: Added a proto file. Also added the sapling definitions.

images/coin/coinZec.png
daira, line 1: I see that the barred Z is transparent. I think that's correct in that it should be white in light mode and black in dark mode (see "Horizontal logos" or "Vertical logos" at https://z.cash/press/ ).

images/walletList/walletListZashi.png
daira, line 1: This is probably not going to work in light mode, because the Zashi logo and name are in white.

rust/apps/zcash/src/pczt/check.rs
```diff
@@ -0,0 +1,341 @@
...
+                                        })
+                                    },
+                                )?;
+                                if target.public_key().serialize().to_vec()
+                                    != input.script_pubkey().clone()
+                                {
+                                    return Err(ZcashError::InvalidPczt(
+                                        "transparent input script pubkey mismatch".to_string(),
+                                    ));
+                                }
```
str4d, line 86: I have not confirmed the alterations to your vendored code, but if `target` here has type `bip32::ExtendedPublicKey<secp256k1::PublicKey>` (like it does inside `zcash_primitives::legacy::keys::AccountPubKey`) then there is a bug here: you are comparing the serialized public key to the `script_pubkey`, but for P2PKH scripts the serialized public key appears in the `script_sig`; only its hash is present in `script_pubkey`.

rust/apps/zcash/src/pczt/check.rs
```diff
@@ -0,0 +1,341 @@
...
+    seed_fingerprint: &[u8; 32],
+    xpub: &AccountPubKey,
+    input: &pczt::transparent::Input,
+) -> Result<(), ZcashError> {
+    let script = Script(input.script_pubkey().clone());
+    match script.address() {
+        Some(TransparentAddress::PublicKeyHash(hash)) => {
+            let pubkey = input.bip32_derivation().keys().find(|pubkey| {
+                return hash[..] == Ripemd160::digest(Sha256::digest(pubkey))[..];
+            });
```
str4d, line 61: Similarly, this check is incorrect:
- `hash` is the hash of the `script_pubkey`
- `input.bip32_derivation()` contains a map from the encoded public key to its derivation.

This follows PSBT semantics (from BIP 174) where the derivation map lists encoded keys; `bip32_derivation` is not a map from `script_pubkey` to derivation path. This is because a P2SH script may contain multiple such keys and require multiple signatures. A P2PKH script happens to only contain one, but for symmetry the PCZT format treats it the same way (like PSBT).

rust/apps/zcash/src/pczt/check.rs
```diff
@@ -0,0 +1,341 @@
...
+    xpub: &AccountPubKey,
+    output: &pczt::transparent::Output,
+) -> Result<(), ZcashError> {
+    let script = Script(output.script_pubkey().clone());
+    match script.address() {
+        Some(TransparentAddress::PublicKeyHash(hash)) => {
+            let pubkey = output
+                .bip32_derivation()
+                .keys()
+                .find(|pubkey| hash[..] == Ripemd160::digest(Sha256::digest(pubkey))[..]);
```
str4d, line 117: This is similarly incorrect.

rust/apps/zcash/src/pczt/check.rs
```diff
@@ -0,0 +1,341 @@
...
+                                        })
+                                    },
+                                )?;
+                                if target.public_key().serialize().to_vec()
+                                    != output.script_pubkey().clone()
+                                {
+                                    return Err(ZcashError::InvalidPczt(
+                                        "transparent output script pubkey mismatch".to_string(),
+                                    ));
+                                }
```
str4d, line 142: This is similarly incorrect.

str4d (changes requested):

rust/apps/zcash/src/pczt/check.rs
```diff
@@ -0,0 +1,343 @@
...
+                                    xpub,
+                                    |acc, n| {
+                                        acc.derive_child(ChildNumber::from(*n)).map_err(|_| {
+                                            ZcashError::InvalidPczt(
+                                                "transparent input bip32 derivation path invalid"
+                                                    .to_string(),
+                                            )
+                                        })
+                                    },
+                                )?;
```
str4d, line 79: Bug: `xpub` is obtained earlier as `ufvk.transparent()`, which means it is already derived at the path `M/44'/133'/account_index'`. However, here the code is applying the transparent inputs entire `bip32_derivation` path to `xpub`, which results in the key at `M/44'/133'/account_index'/44'/133'/account_index'/scope/address_index` (which itself doesn't make sense as you'd be trying to do hardened derivation from a public key, so you're probably just getting an error here immediately).

Instead, the code should check that `bip32_derivation.derivation_path` starts with `44'/133'/account_index'`, and then only derive from `xpub` with the remaining suffix (which should be entirely non-hardened for a valid input, which you'll detect via the existing error check), to obtain (in the usual Zashi tx flow) `M/44'/133'/account_index'/scope/address_index`.

rust/Cargo.toml
```diff
@@ -127,10 +130,18 @@ prost-types = { version = "0.11", default-features = false }
...
+bip32 = { git = "https://github.com/KeystoneHQ/crates.git", rev = "9873e8fd56007d792fa60d6e844fdb75d527c858" }
+f4jumble = { git = "https://github.com/zcash/librustzcash.git", rev = "9407f09208d5574a3ba7bf3e6963741114ba77c2" }
+nonempty = { git = "https://github.com/nuttycom/nonempty.git", rev = "38d37189faecb2a0e3d6adc05aa24e1b93c2483b" }
+orchard = { git = "https://github.com/zcash/orchard.git", rev = "e0cc7ac53ad8c97661b312a8b1c064f4cd3c6629" }
+pczt = { git = "https://github.com/zcash/librustzcash.git", rev = "9407f09208d5574a3ba7bf3e6963741114ba77c2" }
+transparent = { package = "zcash_transparent", git = "https://github.com/zcash/librustzcash.git", rev = "9407f09208d5574a3ba7bf3e6963741114ba77c2" }
+zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "9407f09208d5574a3ba7bf3e6963741114ba77c2" }
+zcash_encoding = { git = "https://github.com/zcash/librustzcash.git", rev = "9407f09208d5574a3ba7bf3e6963741114ba77c2" }
+zcash_keys = { git = "https://github.com/zcash/librustzcash.git", rev = "9407f09208d5574a3ba7bf3e6963741114ba77c2" }
+zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "9407f09208d5574a3ba7bf3e6963741114ba77c2" }
```
str4d, line 147: https://github.com/zcash/librustzcash/pull/1666 has been updated to no longer require a `nonempty` patch (now that it has released a no-std-compatible version):
```suggestion
[patch.crates-io]
bip32 = { git = "https://github.com/KeystoneHQ/crates.git", rev = "9873e8fd56007d792fa60d6e844fdb75d527c858" }
f4jumble = { git = "https://github.com/zcash/librustzcash.git", rev = "ab80376bda49e0af524aea23c1d64b3fe1bbe887" }
orchard = { git = "https://github.com/zcash/orchard.git", rev = "cd3e0901ccac2c630dd7fd03eb496d5030c1bbfe" }
pczt = { git = "https://github.com/zcash/librustzcash.git", rev = "ab80376bda49e0af524aea23c1d64b3fe1bbe887" }
transparent = { package = "zcash_transparent", git = "https://github.com/zcash/librustzcash.git", rev = "ab80376bda49e0af524aea23c1d64b3fe1bbe887" }
zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "ab80376bda49e0af524aea23c1d64b3fe1bbe887" }
zcash_encoding = { git = "https://github.com/zcash/librustzcash.git", rev = "ab80376bda49e0af524aea23c1d64b3fe1bbe887" }
zcash_keys = { git = "https://github.com/zcash/librustzcash.git", rev = "ab80376bda49e0af524aea23c1d64b3fe1bbe887" }
zcash_protocol = { git = "https://github.com/zcash/librustzcash.git", rev = "ab80376bda49e0af524aea23c1d64b3fe1bbe887" }
```

rust/rust_c/src/trng.rs
```diff
@@ -0,0 +1,28 @@
...
+        let len = chunk.len() as u8;
+
+        let ret = unsafe { GenerateTRNGRandomness(chunk.as_mut_ptr(), len) };
+
+        // TODO: Determine how to correctly map error codes from the underlying hardware
+        // into the `getrandom` custom error code space `Error::CUSTOM_START..=u32::MAX`.
+        if ret != 0 {
+            let error = NonZeroU32::new(Error::CUSTOM_START.saturating_add_signed(ret)).unwrap();
+            return Err(Error::from(error));
+        }
```
str4d, line 22: This TODO still needs addressing by someone with access to the TRNG documentation, in particular to confirm that `ret` will never be negative. (Also to confirm that `ret = 0` means success, but I think that's the case as otherwise this function would have been consistently returning an error.)

rust/apps/zcash/src/pczt/parse.rs
```diff
@@ -0,0 +1,573 @@
...
+        let mut result = vec![];
+        temp_memo.iter().for_each(|v| {
+            if *v != 0 {
+                result.push(*v);
+            }
+        });
+        result.reverse();
+
+        return Some(String::from_utf8(result).unwrap());
+    }
```
str4d, line 495: This is subtly incorrect: rather than only stripping trailing `nul` bytes from the memo bytes, it is stripping _every_ `nul` byte from the memo bytes. `nul` is a valid byte in UTF-8 (as part of the NUL character).

If a NUL character cannot be included in rendered memo strings (if Keystone requires C-strings), then it should be replaced with an appropriate Unicode replacement character.

str4d (commented):
Reviewed as of d41a8f232bdbe6008f3676ffb4bf5749cf9e0235. This is not a full review or audit of the PR, just what I had time to look over.

rust/apps/zcash/src/pczt/check.rs
str4d: This is fixed now that the firmware is using `AccountPubKey::derive_pubkey_at_bip32_path`.

rust/apps/zcash/src/pczt/check.rs
str4d: This has been fixed.

rust/apps/zcash/src/pczt/check.rs
str4d: This has been fixed.

rust/rust_c/src/trng.rs
soralit: Just checked the `ret` is asserted in the SE driver level so the error code should always be SUCCESS_CODE(0) here.

rust/apps/zcash/src/pczt/check.rs
```diff
@@ -0,0 +1,339 @@
...
+
+//check output cmx
+fn check_action_output(action: &orchard::pczt::Action) -> Result<(), ZcashError> {
+    action
+        .output()
+        .verify_note_commitment(action.spend())
+        .map_err(|e| {
+            ZcashError::InvalidPczt(alloc::format!("invalid Orchard action cmx: {:?}", e))
+        })?;
+
```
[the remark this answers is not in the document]
str4d: I meant to note that I did later see the comment about CPU management being why this was done later. I'd be interested to know what specifically is the CPU bottleneck.

rust/apps/zcash/src/pczt/parse.rs
soralit: I aim to remove until the first non-zero value but implement a wrong one.

PR #1503, opened 2024-12-13
feat: Use TRNG as a custom `getrandom` backend

str4d:
## Explanation

This enables Rust crates that depend on the `getrandom` crate (either directly or via `rand::rngs::OsRng`) to get secure randomness directly instead of needing it to be passed through a per-callsite C FFI.

## Pre-merge check list

### DIFF
--- a/rust/rust_c/src/bindings.rs
+++ b/rust/rust_c/src/bindings.rs
@@ -1,5 +1,10 @@
+#[cfg(feature = "use-allocator")]
 extern "C" {
     pub fn LogRustMalloc(p: *mut cty::c_void, size: u32);
     pub fn LogRustFree(p: *mut cty::c_void);
     pub fn LogRustPanic(p: *mut cty::c_char);
+}
+
+extern "C" {
+    pub fn GenerateTRNGRandomness(randomness: *mut u8, len: u8) -> i32;
 }
--- a/rust/rust_c/src/trng.rs
+++ b/rust/rust_c/src/trng.rs
@@ -1,0 +1,28 @@
+use core::num::NonZeroU32;
+
+use getrandom::{register_custom_getrandom, Error};
+
+use crate::bindings::GenerateTRNGRandomness;
+
+/// Same as in `keystore.h`.
+const ENTROPY_MAX_LEN: usize = 32;
+
+fn keystone_getrandom(dest: &mut [u8]) -> Result<(), Error> {
+    for chunk in dest.chunks_mut(ENTROPY_MAX_LEN) {
+        // SAFETY: `chunk.len()` is at most `ENTROPY_MAX_LEN` which fits in `u8`.
+        let len = chunk.len() as u8;
+
+        let ret = unsafe { GenerateTRNGRandomness(chunk.as_mut_ptr(), len) };
+
+        // TODO: Determine how to correctly map error codes from the underlying hardware
+        // into the `getrandom` custom error code space `Error::CUSTOM_START..=u32::MAX`.
+        if ret != 0 {
+            let error = NonZeroU32::new(Error::CUSTOM_START.saturating_add_signed(ret)).unwrap();
+            return Err(Error::from(error));
+        }
+    }
+
+    Ok(())
+}
+
+register_custom_getrandom!(keystone_getrandom);
--- a/rust/rust_c/src/zcash/src/lib.rs
+++ b/rust/rust_c/src/zcash/src/lib.rs
@@ -90,24 +90,10 @@ pub extern "C" fn parse_zcash_tx(
 }
 
 #[no_mangle]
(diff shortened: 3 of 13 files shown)


rust/rust_c/src/trng.rs
```diff
@@ -0,0 +1,28 @@
...
+        let len = chunk.len() as u8;
+
+        let ret = unsafe { GenerateTRNGRandomness(chunk.as_mut_ptr(), len) };
+
+        // TODO: Determine how to correctly map error codes from the underlying hardware
+        // into the `getrandom` custom error code space `Error::CUSTOM_START..=u32::MAX`.
+        if ret != 0 {
+            let error = NonZeroU32::new(Error::CUSTOM_START.saturating_add_signed(ret)).unwrap();
+            return Err(Error::from(error));
+        }
```
str4d, line 22: The error codes here come from the underlying calls to `SE_GetDS28S60Rng` and `SE_GetAtecc608bRng`:
https://github.com/KeystoneHQ/keystone3-firmware/blob/08a44673b5fc6275bb71608a50b409849e134a3c/src/managers/keystore.c#L51-L91

I don't have access to their documentation and don't know what error codes they can return. This part of the code needs to be changed to be valid for the possible error codes, _or_ we can ignore the specific error codes and just return a fixed error code inside Rust.

rust/rust_c/src/zcash/src/lib.rs
```diff
@@ -90,24 +90,10 @@ pub extern "C" fn parse_zcash_tx(
...
-pub extern "C" fn sign_zcash_tx(
-    tx: PtrUR,
-    seed: PtrBytes,
-    seed_len: u32,
-    randomness: PtrBytes,
-    randomness_len: u32,
-) -> *mut UREncodeResult {
+pub extern "C" fn sign_zcash_tx(tx: PtrUR, seed: PtrBytes, seed_len: u32) -> *mut UREncodeResult {
     let pczt = extract_ptr_with_type!(tx, ZcashPczt);
     let seed = unsafe { slice::from_raw_parts(seed, seed_len as usize) };
```
daira, line 95:
```suggestion
    let seed = unsafe { slice::from_raw_parts(seed, usize::try_from(seed_len).expect("u32 fits in usize")) };
```
(Needs rustfmt.)

rust/rust_c/src/bindings.rs
```diff
@@ -1,5 +1,10 @@
+#[cfg(feature = "use-allocator")]
 extern "C" {
     pub fn LogRustMalloc(p: *mut cty::c_void, size: u32);
     pub fn LogRustFree(p: *mut cty::c_void);
     pub fn LogRustPanic(p: *mut cty::c_char);
 }
+
+extern "C" {
+    pub fn GenerateTRNGRandomness(randomness: *mut u8, len: u8) -> i32;
```
daira, line 9: Document this function.

rust/rust_c/src/trng.rs
```diff
@@ -0,0 +1,28 @@
...
+
+/// Same as in `keystore.h`.
+const ENTROPY_MAX_LEN: usize = 32;
+
+fn keystone_getrandom(dest: &mut [u8]) -> Result<(), Error> {
+    for chunk in dest.chunks_mut(ENTROPY_MAX_LEN) {
+        // SAFETY: `chunk.len()` is at most `ENTROPY_MAX_LEN` which fits in `u8`.
+        let len = chunk.len() as u8;
+
+        let ret = unsafe { GenerateTRNGRandomness(chunk.as_mut_ptr(), len) };
```
daira, line 15:
```suggestion
        let ret: i32 = unsafe { GenerateTRNGRandomness(chunk.as_mut_ptr(), len) };
```

rust/rust_c/src/trng.rs
```diff
@@ -0,0 +1,28 @@
...
+
+        // TODO: Determine how to correctly map error codes from the underlying hardware
+        // into the `getrandom` custom error code space `Error::CUSTOM_START..=u32::MAX`.
```
daira, line 18: `Error::CUSTOM_START` happens to be 0xC0000000 so this mapping is not 1-1; all error codes `>= 0x3FFFFFFF` will be mapped to `Error::from(u32::MAX)`. It's probably fine.

daira (approved):
Untested ACK with comments.

rust/rust_c/src/trng.rs
str4d: The problem is that the retcode type is `i32` so it can technically go below `Error::CUSTOM_START` if it is ever negative. My suspicion is that it won't actually go negative and this is just an artifact of C default types, but without access to the Secure Element's documentation I cannot know.

soralit: I will do a errcode map later.

PR #1699, opened 2025-04-01
zcash: Reduce size of the signed PCZT before returning it

str4d:
## Explanation

The unsigned PCZT provided by the wallet contains various optional fields that we use to both check the PCZT effects and create the signatures. However, we can assume that the wallet can recover these fields if it needs them, and clear them from the PCZT we return. This should make the animated QR code displayed by the Keystone device smaller and easier to scan.

## Pre-merge check list

## How to test

- Create a Keystone transaction in Zashi.
- Scan the Zashi animated QR code with Keystone.
- Approve the transaction in Keystone.
- Scan the Keystone animated QR code with Zashi.
- Check that the signed transaction completes and is successfully broadcast to the Zcash network.

### DIFF
--- a/rust/apps/zcash/src/lib.rs
+++ b/rust/apps/zcash/src/lib.rs
@@ -7,7 +7,6 @@ pub mod pczt;
 use errors::{Result, ZcashError};
 
 use alloc::{
-    format,
     string::{String, ToString},
     vec::Vec,
 };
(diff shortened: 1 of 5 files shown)


daira:
Re: the CI failure. 
> warning: the feature `error_in_core` has been stable since 1.81.0-nightly and no longer requires an attribute to enable

"has been stable since X.Y.Z-nightly" is kind-of a contradiction in terms. That probably shouldn't even be a warning.

soralit:
CI fails for the test coverage, we will take concern of it later.

Issue #2093, opened 2026-03-26
`rustc` aborts when running `rust_c` tests for cypherpunk firmware


sellout:
Specifically, either
- `zcash::tests::test_aes256_cbc_encrypt` or
- `zcash::tests::test_aes256_cbc_decrypt`

will trigger the abort.

```
$ cargo test --features cypherpunk -p rust_c decrypt
[…]
     Running unittests src/lib.rs (target/debug/deps/rust_c-7f65935fbeb2b12b)

running 5 tests
fatal runtime error: failed to initiate panic, error 3, aborting
error: test failed, to rerun pass `-p rust_c --lib`

Caused by:
  process didn't exit successfully: `/…/rust/target/debug/deps/rust_c-7f65935fbeb2b12b` (signal: 6, SIGABRT: process abort signal)
```
It was introduced by 20a7a90166dfca8d55f6fa84834771684ec6f894, and reverting that commit fixes it
```
$ cargo test --features cypherpunk -p rust_c
[…]
     Running unittests src/lib.rs (target/debug/deps/rust_c-f94c0dae2fa1d29b)

running 5 tests
test common::web_auth::tests::test ... ok
test zcash::tests::test_aes256_cbc_encrypt ... ok
test zcash::tests::test_aes256_cbc_decrypt ... ok
test zcash::tests::test_dep_aes256 ... ok
test common::web_auth::tests::test_calculate_auth_code ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.74s
```
but I don’t see anything obviously bad in that commit.

PR #1951, opened 2025-10-29
Stub testnet support for Zcash in cypherpunk fw

sellout:
## Explanation

This is preliminary work for Zcash multisig support. It keeps the duplicated code more similar between the btc_only and cypherpunk firmwares. Supporting testnet is also a useful feature on its own.

There is no GUI to control this, but changing `MainNet` to `TestNet` in the `HOME_WALLET_CARD_ZEC` entry of `g_walletState` will switch to testnet for Zcash.

Changing the entries for `HOME_WALLET_CARD_BTC` or `HOME_WALLET_CARD_MONERO` will have no effect.

## Pre-merge check list
- [x] PR run build successfully on local machine.
- [x] All unit tests passed locally.

## How to test

Replace `MainNet` with `TestNet` as described above, and use the cypherpunk simulator will perform Zcash operations on testnet.

**NB**: This was developed on top of #1945, but pushed without those commits, since there is no code dependency between them. However, since I needed that PR in order to build the simulator, this might not work without those changes.

PR #2133, opened 2026-04-09
Zcash multi-sig impl for cypherpunk firmware

sellout:
## Explanation

This adds transparent multi-sig support to Zcash in the cypherpunk firmware.

This change is built on #1951, and it contains the zcash app implementation of multi-sig. This doesn’t yet provide a UI.

## Pre-merge check list

__NB__: There are multiple unit tests that are already failing in master. See #2093.

## How to test

``` bash
cargo test --features cypherpunk
```


nuttycom:
@czarcas7ic can you please review once this is updated?

czarcas7ic:
Ping whenever ready and I will review, thanks

rust/apps/zcash/src/pczt/check.rs
```diff
@@ -167,8 +154,31 @@ fn check_transparent_input<P: consensus::Parameters>(
                 }
             }
         }
-        _ => Err(ZcashError::InvalidPczt(
-            "transparent input script pubkey is not a public key hash".to_string(),
+        Some(TransparentAddress::ScriptHash(hash)) => {
```
czarcas7ic, line 157: Given the PR description says this does not yet provide a UI, I understand this may be intentionally scoped to the lower app/signing layer. The issue I am having is that I don’t currently see a way to manually validate the feature end-to-end in the simulator.

The new checker path handles `TransparentAddress::ScriptHash`, but the normal firmware transaction flow still builds review data through `parse_zcash_tx_cypherpunk`, and `parse_transparent_input` still rejects `ScriptHash` inputs. So as this PR stands, I don’t think I can load it in the simulator and verify an actual P2SH multisig spend approval/signing flow (I am sure you are aware of this fact from your PR description, just writing this out for my own sake).

Do you have the other follow on part(s) done already so that I can manually test this on a simulator?

rust/apps/zcash/src/pczt/check.rs
```diff
@@ -167,8 +154,31 @@ fn check_transparent_input<P: consensus::Parameters>(
...
+            match find_my_derivation(seed_fingerprint, input.bip32_derivation()) {
+                None => {
+                    //not my output, pass
+                    Ok(false)
+                }
+                Some((pubkey, derivation)) => {
+                    check_my_pubkey(params, account_index, xpub, pubkey, derivation)?;
+                    // 4: check script hash
+                    if let Some(redeem_script) = input.redeem_script() {
```
czarcas7ic, line 166: I think this P2SH input validation needs one more check. For P2PKH inputs, the scriptPubKey directly commits to the pubkey hash, so checking that the derived wallet pubkey hashes to the pubkey hash closes the loop. For P2SH, the scriptPubKey only commits to the redeem script hash.

In this branch we find a matching wallet derivation from `input.bip32_derivation()`, but that metadata is supplied by the PCZT. If `redeem_script()` is missing, this still returns `Ok(true)`, so we treat the input as ours without proving the P2SH script. If `redeem_script()` is present, we check that it hashes to the P2SH script hash, but we still do not check that our derived pubkey appears in that redeem script.

I think `ScriptHash` inputs should require `redeem_script`, validate that it hashes to the scriptPubKey, and validate that the derived wallet pubkey is actually part of the redeem script before treating the input as ours.

rust/apps/zcash/src/multi_sig/address.rs
```diff
@@ -0,0 +1,223 @@
...
+        bip32::ExtendedPublicKey<secp256k1::PublicKey>,
+        Vec<bip32::ChildNumber>,
+    ),
+    ZcashError,
+> {
+    match xpub {
+        descriptor::Key::Public { .. } => Err(ZcashError::MultiSigWalletAddressCalError(
+            "key is not xpub".to_string(),
+        )),
+        descriptor::Key::Xpub { prefix, key, child } => Ok((*prefix, key.clone(), child.clone())),
```
czarcas7ic, line 87: This is dropping the receive/change address index. `create_multi_sig_address_for_wallet` accepts `change` and `account`, and passes them into `derive_pub_key`, but this branch returns the xpub’s existing child path unchanged and never uses either value.

That means calls for different receive/change indexes can end up building the same multisig script/address. Once the UI flow starts requesting address index 1 or change addresses, those addresses would collide with index 0.

Could we derive the concrete child pubkeys for the requested `change/account` path before building the sortedmulti script, and add coverage proving `(0,0)`, `(0,1)`, and `(1,0)` produce distinct addresses for the supported multisig path?

rust/apps/zcash/src/multi_sig/mod.rs
```diff
@@ -0,0 +1,712 @@
...
+    let script_expression = match multi_sig_type {
+        MultiSigType::P2sh | MultiSigType::P2shTest => vec![ScriptExpression::ScriptHash],
+    };
+
+    let output_descriptors = crypto_account.get_output_descriptors();
+
+    let target_output = output_descriptors
+        .iter()
+        .find(|output| output.get_script_expressions() == script_expression)
+        .unwrap_or(&output_descriptors[0]);
```
czarcas7ic, line 193: I think this should fail instead of falling back to the first descriptor. This code searches for the output descriptor matching the requested `ScriptHash` expression, but if none exists it uses `output_descriptors[0]` and continues parsing that as the multisig xpub source.

That falls through to parsing descriptor 0 even though it did not match the requested script expression. If descriptor 0 otherwise passes the path/key checks, we could silently import the wrong descriptor. It also looks like this would panic if the descriptor list is empty. 

I see this pattern was carried over from the [Bitcoin Keystone multisig path](https://github.com/KeystoneHQ/keystone3-firmware/blob/761243491a40d5fafe06142b8af757e4299da891/rust/apps/bitcoin/src/multi_sig/mod.rs#L193-L213), but I think it is a footgun there too. Can we return an explicit error when the requested script expression is absent?

PR #2134, opened 2026-04-09
Stamp firmware version into signed Zcash PCZTs

### LINKED ISSUE #2148: feat(zcash): stamp firmware version into signed PCZT responses
## Motivation

Complement to the pairing-time version advertisement ([#2147](https://github.com/KeystoneHQ/keystone3-firmware/issues/2147)). Even after a wallet learns the Keystone's version from pairing, we want a second, authoritative signal on the signed response itself. Every signed PCZT should carry a record of the firmware that produced it. This lets wallets gate features like quantum recoverability at the point where the signature actually arrives, not only at pairing.

Full rationale: https://hackmd.io/@valargroup/H1pGeaFh-l

## Proposal

After signing, stamp `global.proprietary["keystone:fw_version"] = [major, minor, build]` (3 raw bytes, matching `SOFTWARE_VERSION_{MAJOR,MINOR,BUILD}` from `src/config/version.h`) into the PCZT via the pczt `Updater` role, before the Redactor strips witness data.

The firmware does not enforce a minimum version, it just reports what it's on. The wallet remains the authority on version policy.

czarcas7ic:
Closes #2148

## Explanation

Every signed PCZT response now carries `global.proprietary["keystone:fw_version"] = [major, minor, build]` (3 raw bytes matching `SOFTWARE_VERSION_{MAJOR,MINOR,BUILD}` from `src/config/version.h`).

This lets wallets detect the Keystone firmware version after scanning the signed QR and gate features accordingly (in our specific case, quantum recoverability). The firmware does not enforce any minimum version, it just reports the version it is on. The wallet is the authority on version policy.

**Changes:**

- `rust/apps/zcash/build.rs` (new) — reads `src/config/version.h` at compile time and generates `KEYSTONE_FW_VERSION` so the Rust const stays in sync with the C header automatically.
- `rust/apps/zcash/src/version.rs` (new) — `Version` type with encode/parse helpers.
- `rust/apps/zcash/src/pczt/sign.rs` — after signing, stamps `keystone:fw_version` into `global.proprietary` via the pczt `Updater` role, before the Redactor strips witness data.
- `src/ui/gui_model/gui_model.c` — simulator-only fix: replaces the synchronous `AsyncExecute` stub with a proper FIFO queue so the cypherpunk simulator doesn't crash during wallet onboarding.
- `src/ui/gui_components/gui_keyboard.c` — simulator-only fix: guards `KbTextAreaHandler` against being called on freed objects during LVGL group cleanup.

## Pre-merge check list

## How to test

```bash
# Unit tests (73 pass, including 4 new stamp-verification tests)
cd rust && cargo test -p app_zcash --features cypherpunk

# Verify build.rs generates the version from version.h
cat target/debug/build/app_zcash-*/out/version_generated.rs

# Simulator build (cypherpunk flavor)
python3 build.py -o simulator -s -t cypherpunk
./build/simulator
```

The stamp can be verified end-to-end by pairing with Zodl, signing a transaction, and inspecting the returned PCZT's `global.proprietary` for the `keystone:fw_version` key.

I also tested this directly with a modified zodl app that rejects a transaction if below the specified version:

https://github.com/user-attachments/assets/f5ed92c8-9a4a-4c99-9680-a037b95c7527

## Docs

I wrote up a short doc on how we will use QR pairing and data added when responding with the signed PCZT in order to achieve version negotiation for the upcoming quantum recoverability feature in zcash

https://hackmd.io/@valargroup/H1pGeaFh-l


src/ui/gui_components/gui_keyboard.c
```diff
@@ -1200,6 +1200,13 @@ char *GuiGetTrueWord(const lv_obj_t *obj, uint16_t btn_id)
 void KbTextAreaHandler(lv_event_t * e)
 {
     lv_event_code_t code = lv_event_get_code(e);
+#ifdef COMPILE_SIMULATOR
```
czarcas7ic, line 1203: The fix here in gui_keyboard and gui_model were needed in order for me to load an existing seed phrase onto the keystone simulator. It would otherwise crash every time.

rust/apps/zcash/src/pczt/sign.rs
```diff
@@ -204,11 +223,104 @@ mod tests {
...
+            .global()
+            .proprietary()
+            .get(PROP_KEY_FW_VERSION)
+            .expect("firmware version stamp must be present");
+        assert_eq!(stamp, &KEYSTONE_FW_VERSION.encode().to_vec());
+    }
+
+    /// Reusable helper: parse the bundled test PCZT and inject a
+    /// test proprietary entry at the global level to verify round-trip.
+    fn pczt_with_min_version(min_version: &[u8]) -> Pczt {
```
nuttycom, line 246: Why isn't this under `#[cfg(test)]`?

rust/apps/zcash/src/pczt/sign.rs
```diff
@@ -204,11 +223,104 @@ mod tests {
...
+        let base = Pczt::parse(&pczt_bytes).unwrap();
+        let min_version = min_version.to_vec();
+        Updater::new(base)
+            .update_global_with(|mut g| {
+                g.set_proprietary("test:min_fw_version".to_string(), min_version);
+            })
+            .finish()
+    }
+
+    fn test_seed() -> Vec<u8> {
```
nuttycom, line 259: Same question here, this seems like it should be under `#[cfg(test)]`

rust/apps/zcash/src/pczt/sign.rs
czarcas7ic, line 207: @nuttycom it is wrapped here, was just out of the diff

rust/apps/zcash/src/pczt/sign.rs
czarcas7ic: See https://github.com/KeystoneHQ/keystone3-firmware/pull/2134/changes#r3096688145

rust/apps/zcash/src/pczt/sign.rs
czarcas7ic: Same comment as here https://github.com/KeystoneHQ/keystone3-firmware/pull/2134/changes#r3096688145

soralit (approved):

PR #2149, opened 2026-04-17
feat(zcash): enable USB signing and account export

Thunnini:
## Explanation

I'm building a Zcash desktop wallet that supports Keystone. On desktop (monitor + webcam), the webcam's resolution and focus limitations make it unreliable to read Keystone's animated QR codes. USB transport is the practical solution for desktop wallets.

Zcash PCZT signing and account export work via QR but are completely blocked via USB — this appears to be an implementation gap rather than an intentional restriction (BTC works over USB in CYPHERPUNK, Monero has the same gap, no security rationale is documented, and the pattern matches how Cardano was initially QR-only before #1422 wired up USB).

This PR was developed with AI assistance (Claude Code).

### Changes

**1. `CheckViewTypeIsAllow` rejects Zcash over USB**
`REMAPVIEW_ZCASH` was missing from both the WEB3 switch and the CYPHERPUNK condition, causing "this view type is not supported" for USB-received PCZTs.

**2. `unlimitHandler` is NULL for ZcashTx**
After the user approves a transaction via USB, `GetSingleUrGenerator()` returned NULL, silently dropping the signed result. A new `sign_zcash_tx_unlimited` uses `FRAGMENT_UNLIMITED_LENGTH` (11000 bytes) to avoid truncating multi-KB signed PCZTs with Orchard proofs (the default 200-byte fragment size would only send the first fragment).

**3. No USB path for Zcash account export (CYPHERPUNK)**
When Zodl wallet sends a `QRHardwareCall` with origin `"zodl"`, the device didn't recognize it. Now `ModelGenerateSyncUR` returns a `ZcashAccounts` UR via existing `get_connect_zcash_wallet_ur()`.

## Pre-merge check list

## How to test

1. **USB PCZT signing**: Send a `zcash-pczt` UR via `CMD_RESOLVE_UR` → device displays transaction → approve → signed PCZT returned via USB
2. **USB account export (CYPHERPUNK)**: Send `QRHardwareCall` with `origin="zodl"` → approve → `ZcashAccounts` UR returned
3. **QR regression**: QR-based Zcash signing and Zodl connection still work
4. **Other chains**: BTC/ETH USB signing unaffected

### DIFF
--- a/rust/rust_c/src/zcash/mod.rs
+++ b/rust/rust_c/src/zcash/mod.rs
@@ -5,7 +5,7 @@ use crate::common::{
     free::Free,
     structs::{SimpleResponse, TransactionCheckResult, TransactionParseResult},
     types::{Ptr, PtrBytes, PtrString, PtrT, PtrUR},
-    ur::{UREncodeResult, FRAGMENT_MAX_LENGTH_DEFAULT},
+    ur::{UREncodeResult, FRAGMENT_MAX_LENGTH_DEFAULT, FRAGMENT_UNLIMITED_LENGTH},
     utils::{convert_c_char, recover_c_char},
 };
 use crate::{extract_array, extract_array_mut};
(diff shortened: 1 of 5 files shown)


soralit (changes requested):
Please follow the KeyDerivation pattern as the following codes. 
Additionally you may need a PR to add Zcash schema to this repo https://github.com/KeystoneHQ/keystone-sdk-rust/blob/master/libs/ur-registry/src/extend/key_derivation_schema.rs

Thunnini:
Hi @soralit, thanks for the review!

Could you share the "following codes" you mentioned? They don't appear to have come through in the comment. I want to make sure I understand the KeyDerivation pattern you'd like me to follow before refactoring.

Also, just to confirm my understanding of the requested change: you'd like the Zcash account export to flow through the existing KeyDerivation path (with a new Zcash schema added to `keystone-sdk-rust`), rather than the early-return branch in `ModelGenerateSyncUR` that I currently use. Is that right?

Happy to open the SDK PR for the schema addition once the approach is clear.

PR #2170, opened 2026-05-24
Add macOS simulator setup

greg0x:
## Summary

Adds a master-based macOS simulator setup branch that makes the cypherpunk simulator easier to build and run locally on macOS.

- Adds `simulator.sh` as the normal entry point for building/running the cypherpunk simulator.
- Makes the runner reopen itself in Terminal.app for QR screen-capture support.
- Skips image regeneration during normal simulator builds so Pillow/PyYAML are not required just to run the simulator.
- Improves Rust tool discovery when Terminal.app starts with a minimal PATH.
- Updates macOS README setup notes.
- Improves simulator QR screen scanning with permission checks, multi-display scanning, scale-aware crop tracking, scanner state reset, and a CoreImage fallback on macOS.

## macOS setup notes

For a normal simulator run:

```bash
brew install sdl2
rustup install $(cat rust-toolchain)
cargo install bindgen-cli
rustup run stable cargo install cbindgen
./simulator.sh
```

For full firmware builds on macOS, install the ARM GCC toolchain too:

```bash
brew install --cask gcc-arm-embedded
rustup target add thumbv7em-none-eabihf
```

`simulator.sh` uses `python3.12` if available, otherwise `python3`. It passes `-s` to `build.py`, so it skips image regeneration. That means `requirements.txt` is only needed when regenerating image assets, not for the normal simulator build.

## Screen Recording / QR scanning

The simulator scans QR codes from the Mac screen, not from the camera. macOS gates this behind Screen Recording permission for the terminal application that launches the simulator.

I previously saw this fail when launching from Ghostty because the screen capture permission was tied to the terminal app. The runner now detects when it is not already running inside Terminal.app and opens a temporary `.command` launcher in Terminal.app. After that, grant Terminal.app permission here:

System Settings > Privacy & Security > Screen Recording > Terminal

After granting permission, rerun `./simulator.sh`. If macOS already had Terminal.app open before the permission change, quitting/reopening Terminal.app may be required before screen capture works.

## Verification

- `bash -n simulator.sh`
- `git diff --check`
- `cargo check --manifest-path rust/Cargo.toml -p sim_qr_reader`
- `/usr/local/bin/python3.12 build.py -t cypherpunk -o simulator -s`

The full simulator build completed successfully on macOS. Existing unrelated C/Rust warnings remain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,8 +115,8 @@ For full details, see [docs/SIMULATOR.md](docs/SIMULATOR.md).
 # Install SDL2
 brew install sdl2
 
-# Create a Python virtual environment.
-# Python 3.12 is known to work; Python 3.14 currently has dependency issues.
+# Optional: create a Python virtual environment.
+# Python 3.12 is known to work for image regeneration dependencies.
 uv venv --python python3.12
 source .venv/bin/activate
 uv pip install -r requirements.txt
@@ -127,6 +127,8 @@ uv pip install -r requirements.txt
 # Run the existing build without rebuilding.
 ./simulator.sh --no-build
 ```
+
+`simulator.sh` skips image regeneration, so Pillow/PyYAML are not required for the normal simulator build. Install `requirements.txt` when you need to regenerate image assets.
 
 ### QR Code Scanning (macOS)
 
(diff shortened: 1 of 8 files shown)


greg0x:
Closing this because it was opened against upstream by mistake. Re-opening against the fork for sharing only.

PR #2184, opened 2026-06-20
Add Ironwood PCZT firmware support

czarcas7ic:
Adds cypherpunk Zcash PCZT parse, check, and sign support for v6/Ironwood while preserving existing Orchard and transparent behavior.

Legacy multi-coins paths now reject v6/Ironwood PCZTs instead of displaying or signing them with incomplete legacy support. Signed responses stamp the firmware version and redact optional PCZT fields, including Ironwood bundle data, before returning QR-sized signing results.

This temporarily pins Zcash dependencies to Ironwood/protocol API revisions until the needed APIs are available from released upstream crates.

Validation:
- `cargo test -p app_zcash --no-default-features --features cypherpunk`
- `cargo test -p app_zcash --no-default-features --features multi_coins`
- `cargo check -p rust_c --no-default-features --features simulator-cypherpunk`
- `cargo check -p rust_c --no-default-features --features simulator-multi-coins`


rust/Cargo.toml
```diff
@@ -120,3 +120,13 @@ getrandom = "0.2"
...
+
+[patch.crates-io]
+orchard = { git = "https://github.com/zcash/orchard", rev = "b2af0a11abe00f59c51258d349c2105fe7a16215" }
+pczt = { git = "https://github.com/valargroup/librustzcash", rev = "dc95dcef33a081b925db551eac8bf6533fff22ed" }
+zcash_address = { git = "https://github.com/valargroup/librustzcash", rev = "dc95dcef33a081b925db551eac8bf6533fff22ed" }
+zcash_encoding = { git = "https://github.com/valargroup/librustzcash", rev = "dc95dcef33a081b925db551eac8bf6533fff22ed" }
+zcash_keys = { git = "https://github.com/valargroup/librustzcash", rev = "dc95dcef33a081b925db551eac8bf6533fff22ed" }
+zcash_primitives = { git = "https://github.com/valargroup/librustzcash", rev = "dc95dcef33a081b925db551eac8bf6533fff22ed" }
+zcash_protocol = { git = "https://github.com/valargroup/librustzcash", rev = "dc95dcef33a081b925db551eac8bf6533fff22ed" }
+zcash_transparent = { git = "https://github.com/valargroup/librustzcash", rev = "dc95dcef33a081b925db551eac8bf6533fff22ed" }
```
czarcas7ic, line 132: Set as patch for now. Will update once respective orchard / librustzcash tags are made.

daira (commented):
Review would be easier if pure refactoring was split into its own commit before the Ironwood support.

rust/apps/zcash/src/pczt/check.rs
nullcopy: See my comment below about this check being duplicated. With the [simplified helper I suggest below](https://github.com/KeystoneHQ/keystone3-firmware/pull/2184/changes/d77e231d8f2ee0fbce7a8450ae8f2bee229bd868#r3461332728), this entire function can be refactored much more simply to just:
```rust
fn check_transparent_input<P: consensus::Parameters>(
    params: &P,
    seed_fingerprint: &[u8; 32],
    account_index: zip32::AccountId,
    xpub: &AccountPubKey,
    input: &transparent::pczt::Input,
) -> Result<bool, ZcashError> {
    let script = input.script_pubkey().clone();

    // p2sh transparent input is not supported yet
    let Some(TransparentAddress::PublicKeyHash(hash)) =
        TransparentAddress::from_script_from_chain(&script)
    else {
        return Err(ZcashError::InvalidPczt(
            "transparent input script pubkey is not a public key hash".to_string(),
        ));
    };

    input
        .bip32_derivation()
        .iter()
        .find(|(_pubkey, derivation)| seed_fingerprint == derivation.seed_fingerprint())
        .map_or(Ok(false), |(pubkey, derivation)| {
            check_transparent_derivation(
                params,
                account_index,
                xpub,
                pubkey,
                derivation,
                "input",
            )?;
            if hash[..] != Ripemd160::digest(Sha256::digest(pubkey))[..] {
                return Err(ZcashError::InvalidPczt(
                    "transparent input script pubkey mismatch".to_string(),
                ));
            }
            Ok(true)
        })
}
```

rust/apps/zcash/src/pczt/mod.rs
```diff
@@ -1,18 +1,76 @@
...
+>(
+    params: &P,
+    seed_fingerprint: &[u8; 32],
+    account_index: zip32::AccountId,
+    xpub: &transparent::keys::AccountPubKey,
+    pubkey: &[u8; 33],
+    derivation: &transparent::pczt::Bip32Derivation,
+    field_label: &str,
+) -> Result<bool, ZcashError> {
+    if seed_fingerprint != derivation.seed_fingerprint() {
```
nullcopy: I'd remove this check and rename this helper just to `check_transparent_derivation()`. Every caller has to check the fingerprint manually before invoking this function, so the only thing this helper checks is that the provided pubkey matches the derived pubkey.

nullcopy: Something like this:
```rust
pub(crate) fn transparent_derivation_matches_selected_account<
    P: zcash_vendor::zcash_protocol::consensus::Parameters,
>(
    params: &P,
    account_index: zip32::AccountId,
    xpub: &transparent::keys::AccountPubKey,
    pubkey: &[u8; 33],
    derivation: &transparent::pczt::Bip32Derivation,
    field_label: &str,
) -> Result<(), ZcashError> {
    let target = xpub
        .derive_pubkey_at_bip32_path(params, account_index, derivation.derivation_path())
        .map_err(|_| {
            ZcashError::InvalidPczt(format!(
                "transparent {field_label} bip32 derivation path invalid"
            ))
        })?;
    if &target.serialize() != pubkey {
        return Err(ZcashError::InvalidPczt(format!(
            "transparent {field_label} script pubkey mismatch"
        )));
    }

    Ok(())
}
```

rust/apps/zcash/src/pczt/check.rs
nullcopy, line 182: I have similar refactoring suggestions [as above](https://github.com/KeystoneHQ/keystone3-firmware/pull/2184/changes/d77e231d8f2ee0fbce7a8450ae8f2bee229bd868#r3460988035)

rust/apps/zcash/src/pczt/check.rs
```diff
@@ -304,9 +281,17 @@ fn check_orchard<P: consensus::Parameters>(
     account_index: zip32::AccountId,
     ufvk: &UnifiedFullViewingKey,
     bundle: &orchard::pczt::Bundle,
+    pool_label: &str,
```
nullcopy: It is confusing for the `pool_label` to be customizable while the bundle type is always `&orchard::pczt::Bundle`. I assume this is a naming conflict between the orchard crate/protocol and the orchard pool... is there anything we can do to make that naming conflict less confusing?

rust/apps/zcash/src/pczt/check.rs
```diff
@@ -336,18 +321,29 @@ fn check_action<P: consensus::Parameters>(
     account_index: zip32::AccountId,
     ufvk: &UnifiedFullViewingKey,
     action: &orchard::pczt::Action,
+    pool_label: &str,
```
nullcopy: Same comment as https://github.com/KeystoneHQ/keystone3-firmware/pull/2184/changes/d77e231d8f2ee0fbce7a8450ae8f2bee229bd868#r3461390068

rust/apps/zcash/src/pczt/check.rs
```diff
@@ -358,6 +354,7 @@ fn check_action_spend<P: consensus::Parameters>(
     account_index: zip32::AccountId,
     fvk: &FullViewingKey,
     spend: &orchard::pczt::Spend,
+    pool_label: &str,
```
nullcopy: Same comment as https://github.com/KeystoneHQ/keystone3-firmware/pull/2184/changes/d77e231d8f2ee0fbce7a8450ae8f2bee229bd868#r3461390068

rust/apps/zcash/src/pczt/check.rs
```diff
@@ -403,15 +402,17 @@ fn is_wallet_orchard_address(fvk: &FullViewingKey, address: &Address) -> bool {
 
 #[cfg(feature = "cypherpunk")]
 // check output cmx and internal-ovk output ownership constraints
-fn check_action_output(
+fn check_action_output<P: consensus::Parameters>(
+    params: &P,
     ufvk: &UnifiedFullViewingKey,
     action: &orchard::pczt::Action,
+    pool_label: &str,
```
nullcopy: Same comment as https://github.com/KeystoneHQ/keystone3-firmware/pull/2184/changes/d77e231d8f2ee0fbce7a8450ae8f2bee229bd868#r3461390068

rust/apps/zcash/src/pczt/check.rs
```diff
@@ -423,59 +424,23 @@ fn check_action_output(
         .transparent()
         .map(|k| orchard::keys::OutgoingViewingKey::from(k.internal_ovk().as_bytes()));
 
-    let mut keys = vec![(Some(external_ovk), false), (Some(internal_ovk), true)];
+    let mut keys = alloc::vec![(Some(external_ovk), false), (Some(internal_ovk), true)];
```
nullcopy: You specifically source from `alloc` here and in several places. Why? I would rather just import `alloc::vec` etc at the top of the file, so we can control in one place which allocator is used, etc

rust/apps/zcash/src/pczt/check.rs
```diff
@@ -423,59 +424,23 @@ fn check_action_output(
...
     for (vk, is_internal_ovk) in keys {
-        if let Some((_, address, _)) = super::parse::decode_output_enc_ciphertext(action, vk.as_ref())?
+        if let Some((_, address, _)) =
+            super::parse::decode_output_enc_ciphertext(action, vk.as_ref())?
         {
             if is_internal_ovk && !is_wallet_orchard_address(fvk, &address) {
-                return Err(ZcashError::InvalidPczt(
-                    "Orchard output was recoverable with an internal OVK but does not belong to this wallet".into(),
-                ));
+                return Err(ZcashError::InvalidPczt(alloc::format!(
```
nullcopy: Same comment about `alloc`

rust/apps/zcash/src/pczt/check.rs
```diff
@@ -423,59 +424,23 @@ fn check_action_output(
...
 }
-
-#[cfg(feature = "cypherpunk")]
-#[cfg(test)]
-mod tests {
-    use super::*;
-    use zcash_vendor::{pczt::Pczt, zcash_protocol::consensus::MAIN_NETWORK};
-
-    #[test]
-    fn test_check_pczt_to_transparent_output() {
```
nullcopy, line 453: Same comment about test removals

rust/apps/zcash/src/pczt/mod.rs
```diff
@@ -1,18 +1,76 @@
...
+use crate::errors::ZcashError;
 
-#[cfg(feature = "cypherpunk")]
-use keystore::algorithms::zcash::sign_message_orchard;
+pub(crate) fn parse_pczt(bytes: &[u8]) -> Result<Pczt, ZcashError> {
+    Pczt::parse(bytes).map_err(|_| ZcashError::InvalidPczt("invalid pczt data".to_string()))
+}
 
-use crate::errors::ZcashError;
+pub(crate) fn validate_supported_pczt(pczt: &Pczt) -> Result<(), ZcashError> {
```
nullcopy, line 20: What is the point of this wrapper?

rust/apps/zcash/src/lib.rs
```diff
@@ -202,316 +215,6 @@ pub fn parse_pczt_multi_coins<P: consensus::Parameters>(
...
 pub fn sign_pczt(pczt: &[u8], seed: &[u8]) -> Result<Vec<u8>> {
-    let pczt =
-        Pczt::parse(pczt).map_err(|_e| ZcashError::InvalidPczt("invalid pczt data".to_string()))?;
+    let pczt = pczt::parse_pczt(pczt)?;
     pczt::sign::sign_pczt(pczt, seed)
 }
-
-#[cfg(feature = "cypherpunk")]
-#[cfg(test)]
-mod tests {
```
nullcopy, line 212: Why are these tests removed. It would be helpful to ensure the purity of this refactor. If the tests are failing after refactor, then that is a red flag

rust/apps/zcash/src/pczt/mod.rs
```diff
@@ -1,18 +1,76 @@
...
+    validate_sapling_bundle_consistency(pczt)?;
 
-pub mod check;
-pub mod parse;
-pub mod sign;
-pub mod structs;
+    Ok(())
+}
+
+fn validate_sapling_bundle_consistency(pczt: &Pczt) -> Result<(), ZcashError> {
```
nullcopy, line 51: The commit message says this function was moved from somewhere. Where did it move from? I can't find the original

rust/apps/zcash/src/lib.rs
```diff
@@ -85,6 +94,26 @@ pub fn check_pczt_multi_coins<P: consensus::Parameters>(
...
+
+    pczt::check::check_pczt_transparent(
+        params,
+        seed_fingerprint,
+        account_index,
+        &account_pubkey,
+        &pczt,
+        true,
+    )?;
+    Ok(())
```
nullcopy, line 111: Why unwrap and then return `Ok(())`? Just return the result of `check_pczt_transparent()` directly

rust/apps/zcash/Cargo.toml
```diff
@@ -17,16 +17,25 @@ bitvec = { version = "1.0.1", default-features = false, features = ["alloc"] }
...
+pczt = { version = "0.7", default-features = false, features = ["orchard", "sapling", "transparent", "zcp-builder"] }
+zcash_primitives = { version = "0.28", default-features = false, features = ["circuits", "test-dependencies", "transparent-inputs"] }
+incrementalmerkletree = { version = "0.8.2", default-features = false }
+shardtree = "0.6.2"
 
 [lints.rust]
-unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage_nightly)'] }
+unexpected_cfgs = { level = "warn", check-cfg = [
+    'cfg(coverage_nightly)',
+    'cfg(zcash_unstable, values("nu6.3"))',
```
nullcopy, line 37: Why are these warns being silenced? They seem like good warnings that should be resolved in the near future, no?

src/ui/gui_chain/multi/gui_zcash.c
nullcopy, line 1: These UI changes seem unrelated to the PCZT handling refactor. This should go into another commit where it is used, or perhaps into it's own commit if this change makes sense on it's own (although it doesn't appear to)

rust/apps/zcash/src/lib.rs
nullcopy: I suspect the tests are removed because of the PCZT version bump? Can we defer the version bump to a later commit, so that we can run the tests and confirm the purity of the refactor? Most (all?) of this refactor seems unrelated to the PCZT interface itself, right?

rust/apps/zcash/src/pczt/check.rs
```diff
@@ -56,6 +58,23 @@ pub fn check_pczt_orchard<P: consensus::Parameters>(
             Ok(())
         })
         .map_err(map_orchard_verifier_error)?;
+    #[cfg(zcash_unstable = "nu6.3")]
+    if should_process_ironwood {
+        verifier
+            .with_ironwood(|bundle| {
+                check_orchard(
```
nullcopy: Why do we `check_orchard` on ironwood? We should probably give this helper a pool-agnostic name

rust/apps/zcash/src/pczt/parse.rs
```diff
@@ -740,79 +897,25 @@ mod tests {
     }
 
     #[test]
-    fn test_decode_pczt_to_p2pkh_transparent_output() {
```
nullcopy, line 743: Please replace these tests with equivalent tests for the new code

rust/apps/zcash/src/lib.rs
nullcopy: And, even after they get removed in the next commit, please replace the tests with equivalent tests for the new code

nullcopy (changes requested):
Flushing comments. Still need to finish reviewing second half of 24f175984d72c3259acda3748f3a5bee02de0630

rust/apps/zcash/src/pczt/check.rs
ValarDragon: Agreed

rust/apps/zcash/src/pczt/mod.rs
nullcopy: nvm, I see it gains meaning in the second commit. For future reference, a noop rename like this just adds noise to a refactor. Next time, just add the wrapped function in the second commit which introduces the need for wrapping

rust/apps/zcash/src/pczt/mod.rs
```diff
@@ -74,3 +99,577 @@ pub(crate) fn transparent_derivation_matches_selected_account<
...
+    zcash_vendor::zip32::AccountId::try_from(account_index)
+        .map(Some)
+        .map_err(|_| unsupported_path())
+}
+
+#[cfg(all(
+    zcash_unstable = "nu6.3",
+    any(feature = "multi_coins", not(feature = "cypherpunk"))
+))]
+pub(crate) fn pczt_requires_cypherpunk_support(pczt: &zcash_vendor::pczt::Pczt) -> bool {
```
nullcopy, line 172: Isn't pczt only used by Zcash, and isn't Zcash only supported in the "cypherpunk" build? So pczt always requires cypherpunk, no?

rust/apps/zcash/src/pczt/mod.rs
```diff
@@ -74,3 +99,577 @@ pub(crate) fn transparent_derivation_matches_selected_account<
...
+    *pczt.global().tx_version() >= 6 || !pczt.ironwood().actions().is_empty()
+}
+
+#[cfg(all(test, feature = "cypherpunk"))]
+pub(crate) mod test_support {
```
nullcopy, line 180: Not reviewing these tests on this pass

rust/apps/zcash/src/pczt/mod.rs
nullcopy: I guess multicoin builds support transparent ZEC? In which case, this check should also include a check for orchard actions (for backwards/fork compatibility)

czarcas7ic: Zcash isn't cypherpunk only. The standard `multi_coins` build supports Zcash too, just for transparent txs. Only shielded Orchard/Ironwood needs cypherpunk.

This guard is needed b/c the transparent only path uses the old pre-NU6.3 sighash and can't handle V6/Ironwood, so the reject_legacy_* helpers use it to turn those PCZTs away instead of signing or showing a partial transaction. Without it, a multi_coins build would run an Ironwood PCZT through the transparent path.

rust/apps/zcash/src/lib.rs
```diff
@@ -134,6 +135,21 @@ fn transparent_account_pubkey_from_xpub(
         .map_err(|e| ZcashError::InvalidDataError(e.to_string()))
 }
 
+#[cfg(feature = "multi_coins")]
+fn reject_legacy_check_unsupported_pczt(pczt: &Pczt) -> Result<()> {
+    #[cfg(zcash_unstable = "nu6.3")]
+    {
+        // The legacy multi-coins check path only verifies transparent data.
+        // Reject V6/Ironwood PCZTs so check, parse, and sign enforce the same boundary.
+        if pczt::pczt_requires_cypherpunk_support(pczt) {
```
nullcopy, line 145: Shouldn't orchard also be rejected? I'm curious how the legacy path got away without an orchard check

rust/apps/zcash/src/lib.rs
czarcas7ic: Resolved this across the three new split commits

rust/apps/zcash/src/pczt/check.rs
czarcas7ic: Fixed here https://github.com/KeystoneHQ/keystone3-firmware/pull/2184/commits/28380dace6717b5d1655135c5538fca6c7ac3bc5

rust/apps/zcash/src/pczt/check.rs
czarcas7ic: Same fix here https://github.com/KeystoneHQ/keystone3-firmware/pull/2184/commits/28380dace6717b5d1655135c5538fca6c7ac3bc5

rust/apps/zcash/src/pczt/check.rs
czarcas7ic: Same fix here https://github.com/KeystoneHQ/keystone3-firmware/pull/2184/commits/28380dace6717b5d1655135c5538fca6c7ac3bc5

rust/apps/zcash/src/pczt/check.rs
czarcas7ic: Same fix here https://github.com/KeystoneHQ/keystone3-firmware/pull/2184/commits/28380dace6717b5d1655135c5538fca6c7ac3bc5

rust/apps/zcash/src/pczt/check.rs
czarcas7ic: We now call `check_shielded_bundle` which takes respective `ShieldedPool` type

rust/apps/zcash/src/pczt/mod.rs
czarcas7ic: Re-split so this should be clearer when introduced

rust/apps/zcash/src/pczt/mod.rs
czarcas7ic: Thanks, addressed here https://github.com/KeystoneHQ/keystone3-firmware/pull/2184/commits/d6e8bbde8fd7cbe4a6a5181002d34cf35bca7af2

rust/apps/zcash/src/pczt/check.rs
czarcas7ic: https://github.com/KeystoneHQ/keystone3-firmware/pull/2184/commits/658f5b57534761555adb2ef83b7169af863637f3

rust/apps/zcash/src/pczt/check.rs
czarcas7ic: https://github.com/KeystoneHQ/keystone3-firmware/pull/2184/commits/658f5b57534761555adb2ef83b7169af863637f3

rust/apps/zcash/src/pczt/check.rs
czarcas7ic: Done here https://github.com/KeystoneHQ/keystone3-firmware/pull/2184/commits/e76f36e32bae0dd5d2656bd58a614b69020dc646

czarcas7ic: Done above https://github.com/KeystoneHQ/keystone3-firmware/pull/2184/commits/e76f36e32bae0dd5d2656bd58a614b69020dc646

rust/apps/zcash/src/pczt/parse.rs
czarcas7ic: https://github.com/KeystoneHQ/keystone3-firmware/pull/2184/commits/612a4a94456f1e2fa906db2f8159707c79432db0

rust/apps/zcash/src/pczt/check.rs
czarcas7ic: Moved to the lib.rs tests. `test_parse_pczt_orchard_decodes_spend_and_change` parses an Orchard PCZT then runs `check_pczt_cypherpunk` on it: https://github.com/valargroup/keystone3-firmware/blob/612a4a94456f1e2fa906db2f8159707c79432db0/rust/apps/zcash/src/lib.rs#L615-L656. The Ironwood and OVK tests exercise `check` too.

rust/apps/zcash/src/pczt/mod.rs
czarcas7ic: From `check.rs`, where it was `validate_sapling_bundle_consistency` with the same body: https://github.com/KeystoneHQ/keystone3-firmware/blob/fa35b62f81af1823e0c3d71352f5817a989e74f1/rust/apps/zcash/src/pczt/check.rs#L23-L42

src/ui/gui_chain/multi/gui_zcash.c
czarcas7ic: IMO its related since this commit handles adding the ironwood PCZT fields that we need to be able to render in the UI right?

rust/apps/zcash/src/lib.rs
czarcas7ic: We currently can't return it directly since `check_pczt_transparent` returns `Result<bool>` now and we don't use the bool here. Can switch to `.map(|_| ())` if you'd rather.

rust/apps/zcash/Cargo.toml
czarcas7ic: This isn’t silencing real warnings, it just tells Rust that `zcash_unstable="nu6.3"` is an intentional cfg used by Ironwood.

Without this, Rust warns on every intentional use and suggests adding this exact `check-cfg` entry.

rust/apps/zcash/src/lib.rs
czarcas7ic: The legacy reject now covers any shielded bundle, not just V6/Ironwood. `pczt_requires_cypherpunk_support` also returns true for Orchard actions and Sapling spends/outputs, so check, parse, and sign all refuse non-transparent PCZTs. https://github.com/valargroup/keystone3-firmware/blob/c0faa647afa0a6a1a0ca4e2bde37ee3df2adc20a/rust/apps/zcash/src/pczt/mod.rs#L164-L177

rust/apps/zcash/Cargo.toml
nullcopy: I understand that, but IMO that's a good warning. Once nu6.3 becomes stable, the config can be removed and the warning will go away, but until that time, it's useful to warn the dev that unstable features are enabled.

Anyways, that's really up to Keystone devs. I just figured I'd call it out

rust/apps/zcash/src/pczt/mod.rs
nullcopy: Can we update the transparent path to use the new NU6.3 sighash? It would be nice to keep zcash support consistent across builds

src/ui/gui_chain/multi/gui_zcash.c
nullcopy: Unless I misunderstood, this commit (the first commit) was supposed to be a pure refactor, and the second commit added Ironwood.

rust/apps/zcash/src/lib.rs
nullcopy: Ah right. No, what you have is fine

rust/apps/zcash/src/pczt/mod.rs
czarcas7ic: https://github.com/KeystoneHQ/keystone3-firmware/pull/2184/changes/29d9e3489ccb58d821c3a8f98ff8d8afa4182756

I still need to manually test this but its late, will do tomorrow. Would be great to get your eyes on it though, thanks

czarcas7ic: @nullcopy I was able to validate v5 and v6. Nothing shows on the Keystone when you are signing, but this is not a new issue. Additionally, multi coins doesn't even have Zodl as an option, so I had to hack this to even test it.

PR #2192, opened 2026-06-24
Zcash upstream

soralit:
## Pre-merge check list


src/ui/gui_chain/multi/gui_zcash.c
```diff
@@ -32,19 +42,21 @@ void GuiSetZcashUrData(URParseResult *urResult, URParseMultiResult *urMultiResul
...
     GetZcashSFP(GetCurrentAccountIndex(), sfp);
 
     PtrT_TransactionParseResult_DisplayPczt parseResult = NULL;
     do {
 #ifdef WEB3_VERSION
-        parseResult = parse_zcash_tx_multi_coins(data, sfp);
+        parseResult = parse_zcash_tx_multi_coins(g_checkedPczt, sfp);
 #endif
 #ifdef CYPHERPUNK_VERSION
         char ufvk[ZCASH_UFVK_MAX_LEN] = {'\0'};
```
nullcopy: Exact-sized buffer does not leave room for a null-terminator as expected by `strcpy_s`. We should be careful to add +1 to all buffer sizes.

nullcopy: This problem is pervasive, because we use `strcpy_s` a lot, but use exact sizes for most of our buffers. I see many are pre-existing. We should make a pass over the codebase to eliminate these.

src/ui/gui_widgets/multi/cypherpunk/gui_zcash_batch_widgets.c
```diff
@@ -0,0 +1,478 @@
...
+}
+
+void GuiZcashBatchWidgetsRefresh(void)
+{
+    if (g_parseResult == NULL || g_parseResult->error_code != 0) {
+        return;
+    }
+
+    ZcashBatchNavBarRefresh();
+    g_currentTransaction = &g_displayZcashBatch->txs->data[g_currentTxIndex];
```
nullcopy, line 411: Need null check for `g_displayZcashBatch`

src/ui/gui_widgets/multi/cypherpunk/gui_zcash_batch_widgets.c
```diff
@@ -0,0 +1,478 @@
...
+                               checkResult->error_message);
+                }
+                free_TransactionCheckResult(checkResult);
+            }
+            return NULL;
+        }
+        free_TransactionCheckResult(checkResult);
+    }
+
+    uint8_t sfp[32];
```
nullcopy: Missing initialization

src/ui/gui_widgets/multi/cypherpunk/gui_zcash_batch_widgets.c
```diff
@@ -0,0 +1,478 @@
...
+        _("scan_qr_code_error_invalid_qrcode"),
+        descText,
+        NULL,
+        _("OK"),
+        WHITE_COLOR_OPA20);
+    lv_obj_add_event_cb(GuiGetHintBoxRightBtn(cont), CloseParseErrorHandler, LV_EVENT_CLICKED, cont);
+    return cont;
+}
+
+static void OnReturnHandler(lv_event_t *e)
```
nullcopy, line 322: USB request ID not cleared on reject. `OnReturnHandler` rejects and returns home but never calls `ClearUSBRequestId()`. `IsZcashBatchUsbMode()` derives solely from `GetCurrentUSParsingRequestID() != USB_REQUEST_IDLE`, so it stays true.. a later QR-scanned batch takes the USB branch (line 176) and pushes the signature over USB instead of showing the QR. The pullout path (line 219) clears it correctly.

src/ui/gui_widgets/multi/cypherpunk/gui_zcash_batch_widgets.c
```diff
@@ -0,0 +1,478 @@
...
+static void HandleClickPreviousBtn(lv_event_t *e);
+static void HandleClickNextBtn(lv_event_t *e);
+static void GuiReturnHome(void);
+static void CloseParseErrorHandler(lv_event_t *e);
+static UREncodeResult *SignZcashBatchInternal(void *data, bool unlimited);
+static bool IsZcashBatchUsbMode(void);
+static void RejectZcashBatchUsbRequest(void);
+static void RespondZcashBatchUsbParseError(const char *errorMessage);
+
+static void FreeCheckedBatch(void)
```
nullcopy, line 67: use-after-free during signing. FreeCheckedBatch() is reachable from the UI task via GuiZcashBatchWidgetsDeInit → ClearPageData while the background task is inside sign_zcash_batch_tx_cypherpunk(g_checkedBatch, …). A 40-PCZT batch makes this window wide.

nullcopy (commented):
Flushing comments

src/ui/gui_chain/multi/gui_zcash.c
czarcas7ic: https://github.com/KeystoneHQ/keystone3-firmware/pull/2237/commits/5f158dd32d3f6e0fd6de09c32045ad0846a4362e

I only solved the UFVK sites, not on the broader codebase.

src/ui/gui_widgets/multi/cypherpunk/gui_zcash_batch_widgets.c
czarcas7ic: https://github.com/KeystoneHQ/keystone3-firmware/pull/2237/commits/c634f1ca455dd7c093764079b83608469ea12435

src/ui/gui_widgets/multi/cypherpunk/gui_zcash_batch_widgets.c
czarcas7ic: https://github.com/KeystoneHQ/keystone3-firmware/pull/2237/commits/56df2725665dc46054d23f01d7fc8a5a5b156aec

src/ui/gui_widgets/multi/cypherpunk/gui_zcash_batch_widgets.c
czarcas7ic: https://github.com/KeystoneHQ/keystone3-firmware/pull/2237/commits/3b8995fa26cd3dc9793151a8d452f9a31073f9a8

src/ui/gui_widgets/multi/cypherpunk/gui_zcash_batch_widgets.c
czarcas7ic: Looked into this,

`RejectZcashBatchUsbRequest()` calls `HandleURResultViaUSBFunc(..., PRS_PARSING_REJECTED)`, whose shared handler sends the response and then sets `g_requestID = REQUEST_ID_IDLE`. The pullout path needs an explicit clear because it sends no response.

src/ui/gui_chain/multi/gui_zcash.c
nullcopy: Yeah, I think broad codebase cleanup is work for another day

rust/apps/zcash/src/pczt/check.rs
```diff
@@ -373,109 +598,70 @@ fn check_action_spend<P: consensus::Parameters>(
...
-        }
-    }
-
-    Ok(())
+    Ok(parsed_to)
 }
 
 #[cfg(feature = "cypherpunk")]
-#[cfg(test)]
-mod tests {
```
nullcopy, line 448: Can we add these tests back?

rust/apps/zcash/src/pczt/mod.rs
```diff
@@ -1,18 +1,993 @@
...
+
+        // The lean v6 shielded sighash (`pczt_ext::shielded_sig_commitment`) implements
+        // the Orchard and Ironwood v6 commitment domains, but NOT the v6 Sapling-spend
+        // domain (which uses a distinct noncompact personalization and omits the
+        // per-spend anchor; ZIP-244). Keystone's cypherpunk signing never produces
+        // Sapling spend authorizations, so reject any v6 PCZT carrying a Sapling spend
+        // rather than committing our Orchard/Ironwood signature to a v5-rules (wrong)
+        // Sapling digest. Sapling OUTPUTS are version-independent in ZIP-244 and remain
+        // supported; v5 transactions are unaffected (the lean v5 Sapling digest matches).
+        if pczt_is_v6(pczt) && !pczt.sapling().spends().is_empty() {
```
nullcopy, line 39: Sapling spends were not supported before. Don't filter on transaction version.

rust/apps/zcash/src/pczt/mod.rs
```diff
@@ -1,18 +1,993 @@
...
+                "Sapling spends are not supported in v6 transactions".to_string(),
+            ));
+        }
+    }
 
-use keystore::algorithms::secp256k1::get_public_key_by_seed;
-use keystore::algorithms::zcash::calculate_seed_fingerprint;
-use zcash_vendor::zcash_keys::keys::UnifiedFullViewingKey;
+    #[cfg(feature = "cypherpunk")]
+    validate_distinct_orchard_protocol_rks(pczt)?;
```
nullcopy, line 47: don't we also need to validate orchard & ironwood pool consistency, same as Sapling? e.g. if ironwood value balance is non-zero, but there are no ironwood actions, that would be an inconsistency to catch here. Same for Orchard pool

rust/apps/zcash/src/pczt/mod.rs
```diff
@@ -1,18 +1,993 @@
...
+}
+
+pub(crate) fn check_transparent_derivation<
+    P: zcash_vendor::zcash_protocol::consensus::Parameters,
+>(
+    params: &P,
+    account_index: zip32::AccountId,
+    xpub: &transparent::keys::AccountPubKey,
+    pubkey: &[u8; 33],
+    derivation: &transparent::pczt::Bip32Derivation,
```
nullcopy, line 132: Doesn't `Bip32Derivation` contain the account_index? Why is `account_index` supplied separately, and what happens if it disagrees with the index provided in the derivation?

src/ui/gui_widgets/gui_transaction_detail_widgets.c
```diff
@@ -190,6 +190,13 @@ void GuiTransactionDetailDeInit()
 //should get error cod here
 void GuiTransactionParseFailed()
 {
+#ifndef BTC_ONLY
```
nullcopy, line 193: I don't get why this is gated from the `BTC_ONLY` fw

rust/apps/zcash/src/pczt/mod.rs
```diff
@@ -1,18 +1,993 @@
...
+#[derive(Clone, Copy)]
+pub(crate) enum ShieldedPool {
+    Orchard,
+    Ironwood,
+}
 
-use crate::errors::ZcashError;
+#[cfg(feature = "cypherpunk")]
+impl ShieldedPool {
+    pub(crate) fn label(self) -> &'static str {
```
nullcopy: Is there a reason not  to just impl the display trait? That seems to be the only way it's used, and would make each call site simpler

rust/apps/zcash/src/pczt/mod.rs
```diff
@@ -1,18 +1,993 @@
...
-pub mod structs;
+/// Returns the supported account declared by a shielded spend derivation that
+/// belongs to this seed. Missing or different seed fingerprints are not ours
+/// and return `None`; matching fingerprints with paths outside
+/// `m/32'/coin_type'/account'` are invalid.
+#[cfg(feature = "cypherpunk")]
+pub(crate) fn matching_seed_supported_orchard_account(
+    seed_fingerprint: &[u8; 32],
+    derivation: Option<&zcash_vendor::orchard::pczt::Zip32Derivation>,
+    coin_type: u32,
```
nullcopy, line 179: Isn't `coin_type` specified by the derivation? What happens if the derivation's `coin_type` disagrees with the provided?

rust/apps/zcash/src/pczt/mod.rs
```diff
@@ -1,18 +1,993 @@
...
+        .map(Some)
+        .map_err(|_| unsupported_path())
+}
+
+/// Returns whether a PCZT carries anything the transparent-only legacy path
+/// cannot handle: a v6+ transaction, or any shielded (Sapling/Orchard/Ironwood)
+/// content. These must be checked, parsed, and signed by the cypherpunk build.
+#[cfg(any(feature = "multi_coins", not(feature = "cypherpunk")))]
+pub(crate) fn pczt_requires_cypherpunk_support(pczt: &zcash_vendor::pczt::Pczt) -> bool {
+    *pczt.global().tx_version() >= 6
```
nullcopy: Don't understand why tx v6 is cypherpunk only. Why can't we handle a transparent-only v6 transaction in a non-cypherpunk build?

rust/apps/zcash/src/pczt/parse.rs
```diff
@@ -41,6 +90,71 @@ fn format_zec_value(value: f64) -> String {
...
+            format_zec_value(fee as f64),
+        ))
+    }
+
+    fn checked_add(total: &mut u64, amount: u64, label: &str) -> Result<(), ZcashError> {
+        *total = total
+            .checked_add(amount)
+            .ok_or_else(|| ZcashError::InvalidPczt(format!("display {label} total overflow")))?;
+        Ok(())
+    }
```
nullcopy, line 145: (nb) Not sure the best place to add this, but IMO the parse step should also detect if transparent inputs are sourced from separate accounts, so the user can be warned about cross-account input ownership correlation

rust/apps/zcash/src/lib.rs
```diff
@@ -74,17 +145,70 @@ pub fn check_pczt_cypherpunk<P: consensus::Parameters>(
...
+        params,
+        seed_fingerprint,
+        account_index,
+        xpub,
+        &pczt,
+        false,
+    )?;
+
+    let pczt = match policy {
+        ShieldedActionPolicy::Single => pczt,
```
nullcopy: Why distinguish between a single sign and a batch? It would be simpler to reason about if the single sign was just a batch of size 1.

nullcopy: For example, why don't we check if the PCZT has my inputs here, like we do for the batch case?

rust/apps/zcash/src/lib.rs
```diff
@@ -101,22 +225,23 @@ pub fn check_pczt_multi_coins<P: consensus::Parameters>(
...
-    )
+#[cfg(feature = "multi_coins")]
+fn reject_legacy_check_unsupported_pczt(pczt: &Pczt) -> Result<()> {
+    {
+        // The legacy multi-coins check path only verifies transparent data. Reject any
+        // shielded (Sapling/Orchard/Ironwood) or V6 PCZT so check, parse, and sign
+        // enforce the same transparent-only boundary.
+        if pczt::pczt_requires_cypherpunk_support(pczt) {
+            return Err(ZcashError::InvalidPczt(
+                "Shielded or V6 PCZTs require cypherpunk checking support".to_string(),
```
nullcopy: Again, I don't understand why V6 tx version affects this

rust/apps/zcash/src/lib.rs
```diff
@@ -101,22 +225,23 @@ pub fn check_pczt_multi_coins<P: consensus::Parameters>(
...
-    pczt::check::check_pczt_transparent(
-        params,
-        seed_fingerprint,
-        account_index,
-        &account_pubkey,
-        &pczt,
-        true,
-    )
+#[cfg(feature = "multi_coins")]
+fn reject_legacy_check_unsupported_pczt(pczt: &Pczt) -> Result<()> {
```
nullcopy: This name is confusing to me... perhaps just `reject_unsupported_pczt()`? V6 transactions are hardly legacy

rust/rust_c/src/zcash/mod.rs
```diff
@@ -156,43 +193,683 @@ pub unsafe extern "C" fn parse_zcash_tx_cypherpunk(
...
+                            }
+                            Err(e) => {
+                                error = Some(UREncodeResult::from(e).c_ptr());
+                                break;
+                            }
+                        }
+                    }
+                    // End the secret's lifetime before response serialization
+                    // and UR encoding, while retaining it across every PCZT.
+                    drop(ask_cache);
```
nullcopy, line 616: Just confirming, does the `ask_cache` zeroize spend key material on drop?

rust/Cargo.toml
```diff
@@ -120,3 +120,17 @@ getrandom = "0.2"
 minicbor = { version = "0.19", features = ["alloc"] }
 zeroize = { version = "1.8.2", default-features = false }
 # third party dependencies end
+
+[patch.crates-io]
+pczt = { git = "https://github.com/zcash/librustzcash", rev = "878db2074ae8ac2682d3e6c61c00f7018b6adc0c" }
```
nullcopy: These could be updated

nullcopy (commented):
I won't block on it, but my biggest objection is the fact that singleton vs batch signing have independent logic. I would strongly prefer if the singleton case was just treated as a batch size of 1, so that we only had to account for one set of validation/signing logic.

rust/apps/zcash/src/lib.rs
czarcas7ic: @nullcopy right now batch rejects transparent and also returns just compact signatures while the singleton returns the full PCZT. If we had more time I would agree that we should just probably get transparent to work in batch and move everyone over to compact signature responses, but at a minimum I don't feel comfortable forcing everyone to only have compact signatures as a response, which was the biggest speedup in all of our work.

czarcas7ic: This is obviously fixable, I am just worried with our timelines that addressing this correctly might do more harm than good.

rust/rust_c/src/zcash/mod.rs
czarcas7ic: Yes, `SpendAuthCache` owns the slot, whose `Drop` zeroizes `ask`

https://github.com/KeystoneHQ/keystone3-firmware/blob/080e3c37da8bb5d863bbd2ba517fc09b375370d8/rust/apps/zcash/src/pczt/sign.rs#L174-L190


We also have a byte-level test verifying the full backing storage is zero

https://github.com/KeystoneHQ/keystone3-firmware/blob/080e3c37da8bb5d863bbd2ba517fc09b375370d8/rust/apps/zcash/src/pczt/sign.rs#L618-L653

rust/apps/zcash/src/lib.rs
nullcopy: Understand the timeline risk. Let's track it as high-priority refactor work for a follow-up release

rust/apps/zcash/src/pczt/parse.rs
czarcas7ic: The current check rejects a same seed transparent input whose account differs from the selected account, so at least as it stands the situation you describe should not be possible.

rust/apps/zcash/src/pczt/mod.rs
czarcas7ic: The derivation path’s coin type is supplied by the caller. The separate value comes from the active network and is the trusted value used to reject mismatched paths.

Rejection when the path coin type differs from trusted coin type:

https://github.com/KeystoneHQ/keystone3-firmware/blob/080e3c37da8bb5d863bbd2ba517fc09b375370d8/rust/apps/zcash/src/pczt/mod.rs#L206-L214

Trusted coin type is supplied from active network param (params):

https://github.com/KeystoneHQ/keystone3-firmware/blob/080e3c37da8bb5d863bbd2ba517fc09b375370d8/rust/apps/zcash/src/pczt/check.rs#L569-L579

And a test that asserts this:

https://github.com/KeystoneHQ/keystone3-firmware/blob/080e3c37da8bb5d863bbd2ba517fc09b375370d8/rust/apps/zcash/src/lib.rs#L1869-L1911

src/ui/gui_widgets/gui_transaction_detail_widgets.c
czarcas7ic: The BTC ONLY firmware doesn't support signing transactions over USB

rust/apps/zcash/src/pczt/mod.rs
czarcas7ic: The PCZT path includes an account number, but that value is supplied by the host. The firmware separately expects account 0 and rejects a path containing any other account.

The firmware supplies account 0:

https://github.com/KeystoneHQ/keystone3-firmware/blob/080e3c37da8bb5d863bbd2ba517fc09b375370d8/src/ui/gui_chain/multi/gui_zcash.c#L332-L350

Our code passes both the expected account and the host provided path into the derivation check:

https://github.com/KeystoneHQ/keystone3-firmware/blob/080e3c37da8bb5d863bbd2ba517fc09b375370d8/rust/apps/zcash/src/pczt/mod.rs#L115-L136

`librustzcash` compares the path’s account number against the expected account and returns an error if they differ:

https://github.com/zcash/librustzcash/blob/878db2074ae8ac2682d3e6c61c00f7018b6adc0c/zcash_transparent/src/keys.rs#L402-L432

This test supplies an account-1 path while account 0 is expected and asserts that the PCZT is rejected:

https://github.com/KeystoneHQ/keystone3-firmware/blob/080e3c37da8bb5d863bbd2ba517fc09b375370d8/rust/apps/zcash/src/lib.rs#L741-L795

rust/apps/zcash/src/pczt/mod.rs
czarcas7ic: Technically on the singleton flow, the V5 flow can support a multiparty PCZT that has Sapling spends. I doubt anyone would hit this, but not filtering would technically reject existing behavior.

rust/apps/zcash/src/pczt/mod.rs
czarcas7ic: https://github.com/KeystoneHQ/keystone3-firmware/pull/2240/changes/83ac1e21afe218984d18a69eef08aeaae118ee5f

rust/apps/zcash/src/lib.rs
czarcas7ic: https://github.com/KeystoneHQ/keystone3-firmware/pull/2240/changes/0f0dc78a2348be3edf09b5c9ea2c98ee4ceb4b78

rust/apps/zcash/src/lib.rs
czarcas7ic: Added the check for has my inputs here though https://github.com/KeystoneHQ/keystone3-firmware/pull/2240/changes/4277d422879d285291e89b39751e1602ae884e81

rust/apps/zcash/src/pczt/mod.rs
czarcas7ic: Empty validation added here https://github.com/KeystoneHQ/keystone3-firmware/pull/2240/changes/5e05a21a8f384377e3746dc8511b5f6d4fd4d39d

rust/apps/zcash/src/pczt/check.rs
czarcas7ic: https://github.com/KeystoneHQ/keystone3-firmware/pull/2240/changes/4bdf2148d12220cdfe4cf1090174c2c183b54dc4

rust/Cargo.toml
czarcas7ic: https://github.com/KeystoneHQ/keystone3-firmware/pull/2240/changes/ef5ad05f68aac6551348dcead10f2078afe83fde

rust/apps/zcash/src/pczt/check.rs
czarcas7ic: https://github.com/KeystoneHQ/keystone3-firmware/pull/2240/changes/75b6955305be65c9f652714d19c854b2d70db8af

rust/apps/zcash/src/pczt/mod.rs
czarcas7ic: Addressed here

https://github.com/KeystoneHQ/keystone3-firmware/pull/2240/changes/ba4ecf288d06d844d9c9f9cc5db5689c8701d590

I also tightened the gate so a V6 PCZT is only accepted with a supported version, version-group, and consensus-branch combination:

https://github.com/KeystoneHQ/keystone3-firmware/pull/2240/changes/0d02c35564aeba4e440f047889d25fc976c196dc

src/ui/gui_widgets/gui_transaction_detail_widgets.c
nullcopy: TIL :+1:

rust/apps/zcash/src/lib.rs
czarcas7ic: Same fix here. V6 itself no longer causes rejection. The transparent only path now rejects shielded content or unsupported transaction formats, while accepting valid transparent only V6 PCZTs:

https://github.com/KeystoneHQ/keystone3-firmware/pull/2240/changes/ba4ecf288d06d844d9c9f9cc5db5689c8701d590

https://github.com/KeystoneHQ/keystone3-firmware/pull/2240/changes/0d02c35564aeba4e440f047889d25fc976c196dc

rust/apps/zcash/src/pczt/mod.rs
nullcopy: Not sure where this comment landed: https://github.com/KeystoneHQ/keystone3-firmware/commit/83ac1e21afe218984d18a69eef08aeaae118ee5f#r193249810

But now that `pool` impls `Display`, we no longer need all the `pool_label` aliases (which was mostly the point of implementing `Display`)

rust/apps/zcash/src/pczt/check.rs
nullcopy: This is a good start, but I think we should make this test stronger in a few ways:
- https://github.com/KeystoneHQ/keystone3-firmware/commit/75b6955305be65c9f652714d19c854b2d70db8af#r193251462
- https://github.com/KeystoneHQ/keystone3-firmware/commit/75b6955305be65c9f652714d19c854b2d70db8af#r193251845
- https://github.com/KeystoneHQ/keystone3-firmware/commit/75b6955305be65c9f652714d19c854b2d70db8af#r193252040

rust/apps/zcash/src/pczt/mod.rs
czarcas7ic: Agreed, done here

https://github.com/KeystoneHQ/keystone3-firmware/pull/2240/changes/9074e08717a57d9780e1eb00edb6727ca8aa5886

rust/apps/zcash/src/pczt/check.rs
czarcas7ic: Updated here https://github.com/KeystoneHQ/keystone3-firmware/pull/2240/changes/2c54cb538ea9c3a94c1b70efb5d3d329bda8ca27

rust/Cargo.toml
```diff
@@ -120,3 +120,8 @@ getrandom = "0.2"
 minicbor = { version = "0.19", features = ["alloc"] }
 zeroize = { version = "1.8.2", default-features = false }
 # third party dependencies end
+
+[patch.crates-io]
+# The published crate does not yet include the PCZT-owned batch messages or
+# firmware version in batch results.
+ur-registry = { git = "https://github.com/KeystoneHQ/keystone-sdk-rust.git", rev = "4131bae5407e31633c01edf9c324094aca69341a" }
```
czarcas7ic: @soralit I believe this needs a tag so we can remove this patch

soralit: Added 1.0.7

PR #2198, opened 2026-07-03
Optimize Zcash migration signing flow

czarcas7ic:
## Summary

Optimizes the Zcash migration signing flow: a 10-note Orchard-to-Ironwood migration that took ~15 minutes end to end (with ~9 minutes of signing) now completes in ~2.5 minutes. The device reviews a migration batch as one split transaction plus a single aggregate migration summary instead of one full page per child, signs the batch with the selected account only, and returns a compact signatures-only response instead of echoing full signed PCZTs.

## What changed

- **Aggregate migration review**: render the split transaction plus one migration summary (totals and per-child input/output/fee) instead of one full page per child. Every migration child is still fully verified on-device (spend ownership, nullifier/rk, cv_net, note commitment, value balance, wallet-owned outputs, and output recoverability); only the display is aggregated.
- **Reviewed-batch fingerprint binding**: the review path records a fingerprint over the validated batch, and the signer refuses to sign any batch whose fingerprint differs, so "the device signs exactly the bytes the user reviewed" is enforced rather than assumed from GUI control flow.
- **Sign wallet-controlled zero-value change spends**: ownership decides signability, never value, so post-NU6.3 restricted bundles (which pair each change output with a fabricated wallet-controlled zero-value spend) extract correctly.
- **Compact signatures-only response**: return the pool, action index, and 64-byte signature per action instead of full signed PCZTs, cutting the response to far fewer QR frames.
- **Output recoverability in the migration summary**: the final commit makes the summary review reject a migration child whose Ironwood output is committed to a wallet address but whose ciphertext no key can recover, matching the guarantee the ordinary per-message review already enforces (a review-vs-sign parity gap caught in review).
- Reset the lock timer after a long batch parse so the device does not auto-lock mid-review; name the ZIP-244 enc_ciphertext slice boundaries as shared constants; run the rust_c Zcash FFI tests in CI (new macOS job) and widen the workflow path filter to zcash_vendor and the rust_c zcash module.

### DIFF
--- a/src/ui/gui_widgets/multi/cypherpunk/gui_zcash_batch_widgets.c
+++ b/src/ui/gui_widgets/multi/cypherpunk/gui_zcash_batch_widgets.c
@@ -382,6 +382,7 @@ static void *GuiParseZcashBatchData(void)
 
 void GuiZcashBatchWidgetsTransactionParseSuccess(void)
 {
+    ClearLockScreenTime();
     g_displayZcashBatch = g_parseResult->data;
     g_txCount = g_displayZcashBatch->txs->size;
     g_currentTxIndex = 0;
@@ -390,6 +391,7 @@ void GuiZcashBatchWidgetsTransactionParseSuccess(void)
 
 void GuiZcashBatchWidgetsTransactionParseFail(void)
 {
+    ClearLockScreenTime();
     printf("GuiZcashBatchWidgetsTransactionParseFail\n");
     if (g_parseResult != NULL) {
         printf("error: %s\n", g_parseResult->error_message);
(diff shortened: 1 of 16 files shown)


src/ui/gui_widgets/multi/cypherpunk/gui_zcash_batch_widgets.c
```diff
@@ -382,6 +382,7 @@ static void *GuiParseZcashBatchData(void)
 
 void GuiZcashBatchWidgetsTransactionParseSuccess(void)
 {
+    ClearLockScreenTime();
```
czarcas7ic, line 385: These were needed for decent UX since, when waiting on the "Loading" screen, the screen would stay alive for the entire "Loading" process, and then the moment the loading was done, the screen would shut off.

PR #2202, opened 2026-07-09
refactor(zcash): checked-PCZT lifecycle — check once, C stores normalized bytes

czarcas7ic:
## Summary

Makes the Zcash check step the single policy preflight for both single and batch PCZTs. It resolves compact fields, applies account and signing-policy checks, and returns normalized checked bytes that the UI retains for display and signing.

This removes repeated policy checks and the UFVK dependency from signing, keeps post-sign verification in memory, and updates the pinned PCZT/Orchard stack needed for field resolution. The QR batch path now runs preflight before rendering.

## Validation

- `app_zcash` and `rust_c` tests for cypherpunk and multi-coins
- Rust coverage and formatting gates
- Cypherpunk and WEB3 simulator builds

### DIFF
--- a/rust/apps/zcash/src/pczt/sign.rs
+++ b/rust/apps/zcash/src/pczt/sign.rs
@@ -547,10 +547,11 @@ mod tests {
         let base_sighash = RoleSigner::new(pczt.clone())
             .expect("Ironwood PCZT signer should initialize")
             .shielded_sighash();
-        // The v6 Ironwood sighash must not commit the anchor, so clearing it
-        // (the elided wire form batch children use) must leave the shielded
-        // sighash unchanged; a client-provided anchor may equally stay on the
-        // wire.
+        // Mirror the wallet's batch redaction: clearing the anchor rebuilds the
+        // anchor-elided request the wallet sends for batch children, with the
+        // full-anchor PCZT as its own oracle. The v6 Ironwood sighash does not
+        // commit the anchor, so the elided form must leave the shielded sighash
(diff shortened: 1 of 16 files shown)


rust/apps/zcash/src/pczt/sign.rs
```diff
@@ -531,19 +547,22 @@ mod tests {
         let base_sighash = RoleSigner::new(pczt.clone())
             .expect("Ironwood PCZT signer should initialize")
             .shielded_sighash();
-        let updated_anchor = orchard::Anchor::from_bytes([6u8; 32]).unwrap();
-        let updated_anchor_pczt = Updater::new(pczt.clone())
-            .set_v6_ironwood_anchor(updated_anchor)
-            .expect("v6 Ironwood anchor should be replaceable before proving")
+        // Mirror the wallet's batch redaction: clearing the anchor rebuilds the
```
czarcas7ic, line 555: This and below is just a test, it will confuse you when reviewing if you didn't know this

rust/rust_c/src/zcash/mod.rs
```diff
@@ -90,7 +91,9 @@ pub unsafe extern "C" fn check_zcash_tx_cypherpunk(
     seed_fingerprint: PtrBytes,
     account_index: u32,
     disabled: bool,
+    checked_pczt: Ptr<Ptr<ZcashCheckedPczt>>,
```
czarcas7ic, line 94: This threw me off but this is C syntax for the output (it looks like an input param here but its not)

soralit (approved):

PR #2204, opened 2026-07-10
Zcash: single-pass batch review and output-recoverability enforcement

czarcas7ic:
The batch review check and parse paths are combined into one pass. `check_and_parse_pczt_shielded` runs the verifier once while recording display rows, so each output is decrypted once. `decode_output_enc_ciphertext` is generalized over the Orchard or Ironwood note-encryption domain selected by the caller.

Output validation now delegates to the same `parse_orchard_output` routine used by display, so a check-only path cannot accept an output that no wallet key can recover. The undecryptable Ironwood-output regression is covered directly.

The signing path now keeps the cached spend authorization key in a single inline slot shared by the Orchard and Ironwood passes. The slot is scrubbed on replacement and drop, and the intermediate spending key is scrubbed after derivation, so spend authorization material is not retained in freed heap storage. Signing behavior is unchanged.

### DIFF
--- a/.github/workflows/rust-zcash-checks.yml
+++ b/.github/workflows/rust-zcash-checks.yml
@@ -2,6 +2,10 @@ on:
   pull_request:
     paths:
       - rust/apps/zcash/**
+      - rust/zcash_vendor/**
+      - rust/rust_c/src/zcash/**
+      - rust/rust_c/src/common/ur.rs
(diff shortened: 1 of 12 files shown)


czarcas7ic:
This includes PR 2203 since I can't stack PRs as an external contributor

soralit (approved):

PR #2216, opened 2026-07-13
Reject duplicate Orchard and Ironwood action rk values

czarcas7ic:
Reject Zcash PCZTs that reuse an Orchard or Ironwood randomized validating key (`rk`) across actions.

Orchard and Ironwood spend authorization signatures cover the same transaction-wide shielded sighash. If two actions share an `rk`, a signature produced for one action is valid for the other. This adds a shared Cypherpunk validation check across both pools before parsing, checking, or signing can proceed.

The regression coverage constructs two same-account Orchard actions with the same `alpha` and `rk`, confirms each action is individually consistent, then verifies all production check and signing paths reject the PCZT. A separate case covers duplicates across Orchard and Ironwood.


rust/apps/zcash/src/pczt/mod.rs
```diff
@@ -42,6 +42,37 @@ pub(crate) fn validate_supported_pczt(pczt: &Pczt) -> Result<(), ZcashError> {
...
+/// Orchard and Ironwood spend authorization signatures cover the same transaction-wide
+/// shielded sighash. If two actions share an `rk`, a signature for either action can be
+/// copied to the other.
+#[cfg(feature = "cypherpunk")]
+fn validate_distinct_orchard_protocol_rks(pczt: &Pczt) -> Result<(), ZcashError> {
+    let actions = pczt
+        .orchard()
+        .actions()
+        .iter()
+        .chain(pczt.ironwood().actions().iter());
```
ValarDragon, line 62: Right, good idea to just take the union of both. Its _technically_ not needed, because orchard/ironwood sighash's differ. But I'd rather be defensive here as honest behavior should not do this.

rust/apps/zcash/src/pczt/mod.rs
```diff
@@ -42,6 +42,37 @@ pub(crate) fn validate_supported_pczt(pczt: &Pczt) -> Result<(), ZcashError> {
...
+
+    for (index, action) in actions.clone().enumerate() {
+        if actions
+            .clone()
```
ValarDragon, line 66: We shouldn't need clones in either spot here. (Would be O(N^2) clones)

If its weird with rust syntax, we can just do a for loop over two indices i, j.

rust/apps/zcash/src/pczt/mod.rs
```diff
@@ -42,6 +42,37 @@ pub(crate) fn validate_supported_pczt(pczt: &Pczt) -> Result<(), ZcashError> {
...
+            .skip(index + 1)
```
ValarDragon, line 67: Lets not do .skip, lets start from the current index, so we save half the comparisons

rust/apps/zcash/src/pczt/mod.rs
ValarDragon: I don't think we need better performance than O(N^2)/2, as we plan to be running this on like 16-20 actions in practice. 

If we did though, then we can run a hashmap on it

ValarDragon (approved):
Safety LGTM, I think we should optimize this a bit though

Fixes a real, long standing risk pointed out by @ebfull

PR #2222, opened 2026-07-14
DNM: Add request-derived Zcash testnet support

czarcas7ic:
## Explanation

Adds Zcash testnet support for receive addresses, wallet connections, and PCZT signing. Signing derives the network from `Global.coin_type`, validates it at the check, parse, and sign boundaries, rejects unsupported or mixed-network requests, and uses network-specific key derivation. The Receive and Connect switches are page-local and default to mainnet, while transaction review shows `Network / Testnet` only for testnet requests.

The pinned PCZT crate does not currently expose `Global.coin_type`, so this uses a contained v1/v2 Postcard prefix decoder covered by compatibility tests. Unsupported PCZT versions fail closed, and the decoder can be removed once the dependency exposes the field directly.

## Pre-merge check list

## How to test

1. Build the cypherpunk firmware or simulator and verify Zcash Receive and Connect default to mainnet after entering or re-entering each page.
2. Select Testnet from each page's More menu and verify the generated testnet address or wallet connection QR refreshes.
3. Review and sign mainnet and testnet single and batch PCZTs. Testnet reviews should show `Network / Testnet`; mainnet reviews should retain the existing layout.
4. Verify an unsupported network coin type and a batch containing mixed networks are rejected before signing.

Local validation included the Zcash, keystore, and Rust FFI test suites, formatting checks, and a full cypherpunk simulator build and link.


rust/apps/zcash/src/pczt/mod.rs
```diff
@@ -4,15 +4,220 @@ pub mod sign;
...
+struct GlobalNetworkPrefix {
+    _tx_version: u32,
+    _version_group_id: u32,
+    _consensus_branch_id: u32,
+    _fallback_lock_time: Option<u32>,
+    _expiry_height: u32,
+    coin_type: u32,
+}
+
+fn decode_global_coin_type(bytes: &[u8]) -> Result<u32, ZcashError> {
```
ValarDragon, line 101: Why are we doing a second layer of parsing for this instead of using our main parsing function?

src/ui/gui_chain/multi/gui_zcash.c
```diff
@@ -54,9 +54,11 @@ void *GuiGetZcashGUIData(void)
         parseResult = parse_zcash_tx_multi_coins(g_checkedPczt, sfp);
 #endif
 #ifdef CYPHERPUNK_VERSION
-        char ufvk[ZCASH_UFVK_MAX_LEN] = {'\0'};
-        GetZcashUFVK(GetCurrentAccountIndex(), ufvk);
-        parseResult = parse_zcash_tx_cypherpunk(g_checkedPczt, ufvk, sfp);
+        char mainnetUfvk[ZCASH_UFVK_MAX_LEN] = {'\0'};
+        char testnetUfvk[ZCASH_UFVK_MAX_LEN] = {'\0'};
```
nullcopy: Pre-existing, but these should be sized `ZCASH_UFVK_MAX_LEN + 1` to account for the null terminator left by `strcpy_s`

src/ui/gui_widgets/multi/gui_standard_receive_widgets.c
```diff
@@ -918,9 +935,13 @@ static void ModelGetAddress(uint32_t index, AddressDataItem_t *item)
 #ifdef CYPHERPUNK_VERSION
     if (g_chainCard == HOME_WALLET_CARD_ZEC) {
         char ufvk[ZCASH_UFVK_MAX_LEN] = {'\0'};
```
nullcopy: Pre-existing, but these should be sized `ZCASH_UFVK_MAX_LEN + 1` to account for the null terminator left by `strcpy_s`

src/ui/gui_widgets/multi/cypherpunk/gui_connect_wallet_widgets.c
```diff
@@ -356,7 +360,11 @@ UREncodeResult *GuiGetZecData(void)
     keys->size = 1;
     char ufvk[384] = {'\0'};
     uint8_t sfp[32];
```
nullcopy, line 362: `ufvk` and `sfp` sizes should not use magic numbers AND they should be sized +1 to allow for null termination

src/ui/gui_widgets/multi/cypherpunk/gui_zcash_batch_widgets.c
```diff
@@ -148,20 +149,23 @@ PtrT_TransactionCheckResult GuiGetZcashBatchCheckResult(void)
     uint8_t sfp[32] = {0};
```
nullcopy, line 180: Pre-existing, but more magic number & size issues

nullcopy (changes requested):
There's several cases of under-sized buffers, because we use `strcpy_s`, which stamps a null terminator on the last byte of the buffer. Relatedly, several buffers have magic sizes. IMO we should try to use defined sizes everywhere.

I flagged a few cases added in this PR, but it looks like a lot of similar issues merged already. We should probably make a pass over the codebase to clean up all of our buffer definitions.

PR #2242, opened 2026-07-22
Preserve PCZT v1 encoding and Zcash account firmware version

czarcas7ic:
Preserves the incoming PCZT wire version through singleton normalization and signing, so wallets that send PCZT v1 receive a v1 response instead of an incompatible v2 payload. Batch requests intentionally remain v2 because their signability digest is bound to the batch serializer's normalized bytes.

Also updates `ur-registry` and `ur-parse-lib` to their released 1.0.8 versions, removing the temporary git patch. That release encodes the firmware version already supplied in the Zcash account connection response. The new field is optional and encoded after the existing account fields, so older Vizor and Zodl builds continue to decode the response and simply ignore the version. Adds regression coverage for both compatibility paths and release notes for the PCZT v1 change.


rust/Cargo.toml
```diff
@@ -119,4 +119,8 @@ blake2b_simd = { version = "1.0.2", default-features = false }
 getrandom = "0.2"
 minicbor = { version = "0.19", features = ["alloc"] }
 zeroize = { version = "1.8.2", default-features = false }
-# third party dependencies end
\ No newline at end of file
+# third party dependencies end
+
+[patch.crates-io]
+ur-registry = { git = "https://github.com/valargroup/keystone-sdk-rust.git", rev = "abc72f085f4c42e0ad048c60c1102f24a16e1f68" }
+ur-parse-lib = { git = "https://github.com/valargroup/keystone-sdk-rust.git", rev = "abc72f085f4c42e0ad048c60c1102f24a16e1f68" }
```
czarcas7ic: Change when we tag rust sdk

soralit (dismissed):

czarcas7ic:
@soralit let me know when the next rust sdk is tagged and ill update this PR, thanks!

rust/Cargo.toml
czarcas7ic: https://github.com/KeystoneHQ/keystone3-firmware/pull/2242/commits/fa0add39ba52673ad8597370c2fa81014cfae4af

PR #2243, opened 2026-07-24
Fix repeated Zcash wallet preparation on unlock

czarcas7ic:
## Explanation

On device builds, `IsHexStringWithLen(value, 0)` selected a 4096-byte maximum and then called `strnlen_s` with 4097. The firmware safeclib limit is 4096, so the call returned zero and every cached encrypted Zcash UFVK was treated as invalid. That caused wallet public information to be regenerated on every unlock even though the saved record and ciphertext were valid.

Keep the variable-length scan within the safeclib limit, reject input that is not terminated within that bound, and preserve exact-length validation for existing callers. Update the English and Chinese Unreleased changelogs for the user-visible fix.

## Pre-merge check list

- [x] Focused safeclib regression cases passed locally.

## How to test

1. Set up a Cypherpunk wallet that supports Zcash and unlock it once.
2. Lock and unlock the device twice more.
3. Confirm the device no longer shows the wallet setup and extended-public-key generation screens on each unlock.


nullcopy (approved):
Good find and good fix
utACK dd0341996459a1a98b76a7f500ab01881178ee11

soralit (approved):

PR #2248, opened 2026-07-28
feat(zcash): tag outputs to the wallet's own address as Mine

### LINKED ISSUE #2247: Zcash address not recognized as my own
When sending to myself, the transaction review screen does not label the destination address as 'mine', even though it is.

Looking into the code, it appears the address is recognized as my own, but the 'is_mine' label is not attached for external addresses. This is a bug. I will fix ASAP.

nullcopy:
The review screen previously computed shielded output ownership via the IVK scope check and then discarded the external-scope result, so an output paying the wallet's own external Orchard or Ironwood receiver rendered like a payment to a foreign address. Propagate the ownership flag through ParsedTo and DisplayTo and render a "Mine" tag on output rows, with the existing "Change" tag taking precedence for internal-scope outputs.

Closes https://github.com/KeystoneHQ/keystone3-firmware/issues/2247

### DIFF
--- a/rust/apps/zcash/src/lib.rs
+++ b/rust/apps/zcash/src/lib.rs
@@ -526,6 +526,7 @@ impl BatchMigrationSummary {
                         true,
                         false,
                         None,
+                        true,
                     )
                 })
                 .collect(),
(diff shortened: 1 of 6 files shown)


czarcas7ic (approved):
This change checks out to me, thanks

PR #2259, opened 2026-08-04
Bound Zcash batch signing resources

czarcas7ic:
Compact Zcash batch PCZTs expand when firmware restores redacted Orchard fields, while signed-response construction also retains one spend authorization signature per action. Oversized batches could therefore reach memory-intensive processing without a single enforceable resource contract.

This replaces the independent wire-size and PCZT-count limits with one aggregate pre-resolution budget: the request ID, canonical batch serialization using PCZT encoding v2, the exact 8-byte standalone header per PCZT, and worst-case 645-byte growth per Orchard action must fit the existing 512 KiB internal ceiling. The encoded envelope is bounded before PCZT parsing, and the aggregate estimate is checked before field resolution.

A separate ceiling of 96 Orchard actions across the Orchard and Ironwood value pools remains for response construction because 128-action hardware tests reached an SRAM allocation failure while 96-action batches have not. The 512 KiB request budget and 96-action response cap are intentionally independent; physical-device validation of this exact revision remains outstanding.


nullcopy (changes requested):
NACK https://github.com/KeystoneHQ/keystone3-firmware/pull/2259/commits/1f01865374d57e947644936bfd58b6aef396b976

As discussed in dm, the more appropriate change would be to refactor away from a PCZT-count limit, and instead use an action-count limit for batch sizes. This would not be urgent, though, because an action limit can be implemented easily on the client side.

On ZODL, we are limiting our batches to contain no more than 96 actions (regardless of tx composition), and we have not seen any OOM issues yet. This can be implemented by any wallet today, without a Keystone firmware change.

docs/protocols/ur_registrys/zcash.md
```diff
@@ -62,9 +62,14 @@ compact response uses `zcash-batch-sig-result`, echoes the request id, and
 carries the PCZT-owned response in its own opaque `data` field. It also reports
 the signing firmware version once for the entire response.
 
-Batch version 1 is supported by cypherpunk firmware and currently accepts up to
-40 PCZTs. The encoded batch data and request id together, and the canonical PCZT
-payloads after decoding, must each fit within 512 KiB. The operation is atomic.
+Batch version 1 is supported by cypherpunk firmware. Its resource limits are
+128 KiB for the encoded batch data plus request id and 96 total Orchard and
+Ironwood actions. Together they form a sender-checkable contract that guarantees
```
nullcopy: `s/Orchard and Ironwood actions/Orchard actions/`

protocol instead of pool

docs/protocols/ur_registrys/zcash.md
```diff
@@ -62,9 +62,14 @@ compact response uses `zcash-batch-sig-result`, echoes the request id, and
 carries the PCZT-owned response in its own opaque `data` field. It also reports
 the signing firmware version once for the entire response.
 
-Batch version 1 is supported by cypherpunk firmware and currently accepts up to
-40 PCZTs. The encoded batch data and request id together, and the canonical PCZT
-payloads after decoding, must each fit within 512 KiB. The operation is atomic.
+Batch version 1 is supported by cypherpunk firmware. Its resource limits are
+128 KiB for the encoded batch data plus request id and 96 total Orchard and
```
nullcopy: What is the rationale for the 128KiB limit? If a client submits a 512KiB batch of fully transparent PCZTs (no per-action growth to account for), that should be ok, right?

rust/apps/zcash/src/lib.rs
```diff
@@ -36,6 +36,15 @@ use zcash_vendor::sha2::{Digest, Sha256};
 #[cfg(feature = "cypherpunk")]
 use zcash_vendor::zcash_protocol::consensus::NetworkConstants;
 
+/// Maximum serialized growth of one compact Orchard or Ironwood action when
```
nullcopy:
```suggestion
/// Maximum serialized growth of one compact Orchard action when
```

rust/rust_c/src/zcash/mod.rs
```diff
@@ -238,9 +238,9 @@ fn validate_zcash_batch_payloads(payloads: &[Vec<u8>]) -> Result<(), RustCError>
             )));
         }
         total_payload_bytes = total_payload_bytes.saturating_add(payload.len());
-        if total_payload_bytes > ZCASH_BATCH_MAX_TOTAL_BYTES {
+        if total_payload_bytes > ZCASH_BATCH_MAX_CANONICAL_PAYLOAD_BYTES {
```
nullcopy: This is too late to be performing the size check. At this point, the batch is already in memory, and any damage is done.

rust/rust_c/src/zcash/mod.rs
```diff
@@ -313,22 +318,55 @@ fn validate_zcash_batch_request_count(data: &[u8]) -> Result<(), RustCError> {
...
 
     Ok(())
 }
 
-/// Parses the bounded outer registry into the PCZT crate's batch request.
+/// Rejects action counts whose decoded allocation and resolution cost cannot
+/// be bounded by the compact wire size alone.
+#[cfg(feature = "cypherpunk")]
+fn validate_zcash_batch_action_count(batch: &BatchSignRequest) -> Result<(), RustCError> {
+    let total_actions = batch.pczts().iter().fold(0usize, |total, pczt| {
```
nullcopy: (nit) [reduce](https://doc.rust-lang.org/core/iter/trait.Iterator.html#method.reduce) would be cleaner than `fold` here

rust/rust_c/src/zcash/mod.rs
```diff
@@ -37,12 +37,18 @@ use zcash_vendor::pczt::roles::signer::batch::{BatchSignRequest, BatchSignRespon
...
+const ZCASH_BATCH_MAX_WIRE_BYTES: usize = 128 * 1024;
 #[cfg(feature = "cypherpunk")]
-const ZCASH_BATCH_MAX_TOTAL_BYTES: usize = 512 * 1024;
+const ZCASH_BATCH_MAX_TOTAL_ACTIONS: usize = 96;
+#[cfg(feature = "cypherpunk")]
+const ZCASH_BATCH_MAX_NORMALIZED_BYTES: usize = 512 * 1024;
+#[cfg(feature = "cypherpunk")]
+const ZCASH_BATCH_MAX_CANONICAL_PAYLOAD_BYTES: usize = ZCASH_BATCH_MAX_NORMALIZED_BYTES;
+#[cfg(feature = "cypherpunk")]
+const ZCASH_BATCH_MAX_V1_TO_V2_FIXED_GROWTH_PER_PCZT: usize = 7;
```
nullcopy: I am unable to corroborate this number. How did you arrive at 7?

rust/apps/zcash/src/lib.rs
```diff
@@ -36,6 +36,15 @@ use zcash_vendor::sha2::{Digest, Sha256};
...
+/// [`Pczt::resolve_fields`] restores `cv_net` (32 bytes), `cmx` (32 bytes),
+/// and the full ciphertext representation (581 bytes).
```
nullcopy:
```suggestion
/// [`Pczt::resolve_fields`] restores redacted fields.
```

rust/rust_c/src/zcash/mod.rs
```diff
@@ -1053,26 +1078,65 @@ mod tests {
...
+        )
+        .unwrap();
+
+        validate_zcash_batch_action_total(ZCASH_BATCH_MAX_TOTAL_ACTIONS).unwrap();
+        assert!(validate_zcash_batch_action_total(ZCASH_BATCH_MAX_TOTAL_ACTIONS + 1).is_err());
+    }
+
+    #[cfg(feature = "cypherpunk")]
+    #[test]
+    fn test_zcash_batch_wire_contract_fits_normalized_limit() {
```
nullcopy: This test doesn't actually test what we need it to

rust/rust_c/src/zcash/mod.rs
```diff
@@ -37,12 +37,18 @@ use zcash_vendor::pczt::roles::signer::batch::{BatchSignRequest, BatchSignRespon
 use zcash_vendor::{pczt::Pczt, zcash_protocol::consensus::MainNetwork};
 use zeroize::Zeroize;
 
-// Cap both per-PCZT overhead and variable-size payload data to leave headroom
-// in shared device memory while processing a batch.
+// The wire and action limits are the sender-facing contract. Together they
+// guarantee that field resolution fits the separate normalized limit.
 #[cfg(feature = "cypherpunk")]
-const ZCASH_BATCH_MAX_PCZTS: usize = 40;
+const ZCASH_BATCH_MAX_WIRE_BYTES: usize = 128 * 1024;
```
nullcopy: (nit) I understand what each of these constants represent, but some comments describing each would go a long way for a future reader

nullcopy (changes requested):
I'm still not sure this is the correct contract to enforce. Let's agree on the contract first, and then my comments about how it's enforced may or may not go away.

The only thing I can say for certain, is that we have never seen a failure when constructing batches of shielded-only transactions with <= 96 actions. We sent at most 32 PCZTs (3 actions each). For now, lets back out a max batch size from that, and use that as our bound. Anything else is an over-constraint.

So,  `MAX_BATCH_SIZE` would simply be:
```
  96 x MAX_ACTION_SIZE # need to figure out the size from the actions we sent
+ 32 x PER_TX_OVERHEAD # need to figure this out too
```

And then the only constraint Keystone has to enforce, is that the max resolved size of the batch doesn't exceed that number. So throw an error if:
```
  batch.bytes().len()                           # actual wire size
+ batch.actions().len() x MAX_ACTION_SIZE       # action overhead
+ batch.transactions().len() x PER_TX_OVERHEAD  # tx encoding overhead
>= MAX_BATCH_SIZE
```
(pardon my pseudo code)

This is the contract the client should respect when constructing a batch. We should do away with max wire size, PCZT count checks, and all the other growth checking, IMO.

docs/protocols/ur_registrys/zcash.md
czarcas7ic: Was an initial best guess assuming the worst case for resolved fields, but your point makes sense and that is a better design, changed.

rust/rust_c/src/zcash/mod.rs
czarcas7ic: Late check replaced

rust/rust_c/src/zcash/mod.rs
czarcas7ic: Was an estimation from testing various v1 to v2  conversions, but is no longer needed

rust/rust_c/src/zcash/mod.rs
czarcas7ic: Comments added

rust/rust_c/src/zcash/mod.rs
czarcas7ic: New test is renamed `test_zcash_batch_resource_budget_charges_batch_pczts_and_actions`, am leaving unresolved until you acknowledge this was addressed to your satisfaction.

czarcas7ic:
Am testing this on ForgeBox before pinging for next review (feel free to add feedback as desired though)

rust/rust_c/src/zcash/mod.rs
```diff
@@ -276,59 +264,113 @@ fn validate_zcash_batch(batch: &BatchSignRequest) -> Result<Vec<Vec<u8>>, RustCE
...
+#[cfg(feature = "cypherpunk")]
+fn estimate_zcash_batch_resolved_bytes(
+    request_id: &[u8],
+    batch: &BatchSignRequest,
+) -> Result<usize, RustCError> {
+    let canonical_bytes = batch
+        .serialize()
+        .map_err(|e| RustCError::InvalidData(format!("encode PCZT batch request: {e:?}")))?
+        .len();
+    let per_pczt_overhead = batch
```
nullcopy:
```suggestion
    let pczt_encoding_overhead = batch
```

non-blocking/nit-picking 😅

rust/rust_c/src/zcash/mod.rs
```diff
@@ -276,59 +264,113 @@ fn validate_zcash_batch(batch: &BatchSignRequest) -> Result<Vec<Vec<u8>>, RustCE
...
+        .checked_mul(app_zcash::COMPACT_PCZT_MAX_RESOLVED_ACTION_GROWTH)
+        .ok_or_else(|| {
+            RustCError::UnsupportedTransaction("Zcash batch size overflow".to_string())
+        })?;
+
+    request_id
+        .len()
+        .checked_add(canonical_bytes)
+        .and_then(|bytes| bytes.checked_add(per_pczt_overhead))
+        .and_then(|bytes| bytes.checked_add(resolved_action_growth))
```
nullcopy, line 348: This tests for a literal integer overflow, right? We should check that it doesn't exceed `ZCASH_BATCH_MAX_RESOLVED_BYTES`

nullcopy (changes requested):
I like the structure of this a lot more! I think you're missing the actual max size check though

rust/rust_c/src/zcash/mod.rs
czarcas7ic: Yeah those `checked_add` calls are only checking for integer overflow. The actual max size check is directly below in `validate_zcash_batch_resource_budget`, which rejects anything over `ZCASH_BATCH_MAX_RESOLVED_BYTES`:

https://github.com/KeystoneHQ/keystone3-firmware/blob/bac6cb5914a4e0d21f6d195df8fd59a3f5c07664/rust/rust_c/src/zcash/mod.rs#L354-L364

The boundary test also accepts exactly the limit and rejects one byte over:

https://github.com/KeystoneHQ/keystone3-firmware/blob/bac6cb5914a4e0d21f6d195df8fd59a3f5c07664/rust/rust_c/src/zcash/mod.rs#L1122-L1147

Should we just fold them in together?

nullcopy: Ahhh, there it is 🤦

Yeah, unless there's a reason to split them apart, I'd be in favor of folding them together

nullcopy (approved):
utACK c2196b2490436d62b1155e80202c475df758641f

